PR #87, opened 2022-10-18
Update pipenv commands and add requirements tox run

### LINKED ISSUE #73: Pipenv deprecated the lock -r [-d], and replaced it with requirements [--dev].
Pipenv deprecated the `lock -r [-d]`, and replaced it with `requirements [--dev]`.

It requires
- Update the Makefile
- Run the `make requirements`

Also, we don't test if the Pipenv* was updated during the commit.
It's necessary to add a check in the `tox.ini`

See the comment above for more context and details
- https://github.com/kaprien/tuf-repository-service-api/pull/117#pullrequestreview-1133857823

MVrachev:
Fixes #73

The changes in this pr are as follows:
1. update the `pipenv` command in `Makefile`
2. update the `requirements` files content
3. add a tox ci test where we check if there is a new dependency or version of a dependency that should be added to any of the requirements files.

PS: These changes in this pr were first discussed here: https://github.com/vmware/repository-service-tuf-api/pull/124

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py39,py310,lint,test
+envlist = py39,py310,lint,requirements,test
 
 [flake8]
 exclude = ownca/__init__.py,venv,.venv,settings.py,.git,.tox,dist,docs,*lib/python*,*egg,build,tools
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -30,6 +30,17 @@ commands =
     coverage xml -i
     coverage report
 
+[testenv:requirements]
```
kairoaraujo, line 33: Please, add the requirements in the tox envlist (line 2 of this file)

```ini
[tox]
envlist = py39,py310,requirements,lint,test
```

kairoaraujo (changes requested):

tox.ini
MVrachev: Done.

kairoaraujo (approved):

PR #116, opened 2022-10-25
build: add workflow for updating python dependencies

rdimitrov:
The following PR adds a dedicated workflow for updating Python dependencies.

Details:
* Runs daily
* If another one is already opened, it doesn't create a new one, instead, it updates it.

Related to https://github.com/vmware/repository-service-tuf/issues/84

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>


kairoaraujo (approved):
LGTM

PR #123, opened 2022-11-01
Make API docker image public in docker-compose.yml

### LINKED ISSUE #55: Make sure there is not undesirable changes in docker-compose.yml
Uncomment:
https://github.com/kaprien/kaprien-repo-worker/blob/945ea61bdde31ddfb96dfa83abccd628cdefbfaf/docker-compose.yml#L27
Comment:
https://github.com/kaprien/kaprien-repo-worker/blob/945ea61bdde31ddfb96dfa83abccd628cdefbfaf/docker-compose.yml#L28

MVrachev:
Make API docker image public in docker-compose.yml.

Close #55 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+    - name: Check if any local image is used in docker-compose.yml
+      run: |
+        if [[ "$(egrep -w 'image:\s+repository-service-tuf-api|image:\s+repository-service-tuf-worker' docker-compose.yml -c)" -ne "0" ]]; then echo "Local image has been used in docker-compose.yml" && exit 1; fi
+
     - uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984
       with:
         python-version: ${{ matrix.python-versions }}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,8 +23,7 @@ services:
     restart: always
 
   repository-service-tuf-api:
-    #image: ghcr.io/vmware/repository-service-tuf-api:dev
-    image: repository-service-tuf-api:dev
+    image: ghcr.io/vmware/repository-service-tuf-api:dev
     volumes:
       - repository-service-tuf-api-data:/data
     ports:


kairoaraujo:
Do you think it is possible to have an automatic check if the correct entry is in the file when we have PR?

rdimitrov:
> Do you think it is possible to have an automatic check if the correct entry is in the file when we have PR?

Maybe having something like this in our CI?
 
`egrep -w 'image: repository-service-tuf-api|image: repository-service-tuf-worker|image: repository-service-tuf-cli' docker-compose.yml`

.github/workflows/ci.yml
```diff
@@ -15,6 +15,11 @@ jobs:
         python-versions: [ "3.9", "3.10", ]
 
     steps:
+    - name: Check if any local image is used in docker-compose.yml
+      run: |
+        egrep -w 'image: repository-service-tuf-api|image: repository-service-tuf-worker|image: repository-service-tuf-cli' ../../docker-compose.yml >> $GITHUB_OUTPUT
+        if [ "$GITHUB_OUTPUT" != "" ]; echo "Local image has been used in docker-compose.yml" exit 1; fi
+
```
rdimitrov, line 22: This step should be executed after we checkout the code

MVrachev:
I am currently investigating why those CI workflow runs fail when locally they succeed.
Sorry for the many force pushes.

rdimitrov:
The reason is that the line is still there even though you've commented it 👍 

Let's try to change the search pattern in a way that works for us or if we cannot come up with a workaround maybe decide to not leave it commented 👍

MVrachev:
> The reason is that the line is still there even though you've commented it 👍

Maybe you have saw a previous version of the pr, but I have removed the comment a while ago.

The real problem it seems is with bash variables.
Because I wanted to make it more readable I wanted to move the result of the `egrep` command in a bash variable by trying maany things.
It seems that for some reason I had problems instantiating a variable because I tried what you (@rdimitrov) suggested but before I moved it to a one-liner it didn't work.

Thanks for the help @rdimitrov! I am happy to see it solved.

rdimitrov:
> > The reason is that the line is still there even though you've commented it 👍
> 
> Maybe you have saw a previous version of the pr, but I have removed the comment a while ago.
> 
> The real problem it seems is with bash variables. Because I wanted to make it more readable I wanted to move the result of the `egrep` command in a bash variable by trying maany things. It seems that for some reason I had problems instantiating a variable because I tried what you (@rdimitrov) suggested but before I moved it to a one-liner it didn't work.
> 
> Thanks for the help @rdimitrov! I am happy to see it solved.

Glad I helped! 👍 

I noticed that you've discarded the message you previously had in place, so feel free to update what we output in case the image is found 👍

.github/workflows/ci.yml
```diff
@@ -16,6 +16,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+    - name: Check if any local image is used in docker-compose.yml
+      run: |
+        if [[ "$(egrep -w 'image: repository-service-tuf-api|image: repository-service-tuf-worker|image: repository-service-tuf-cli' docker-compose.yml -c)" -ne "0" ]]; then echo "Found" && exit 1; fi
```
kairoaraujo, line 22:
```suggestion
        egrep -w 'image:\s+repository-service-tuf-api|image:\s+repository-service-tuf-worker' docker-compose.yml
        if [[ $rc -eq "0" ]]; then \
          echo "Local image has been used in docker-compose.yml" \
          exit 1 \
        fi
```

- The `repository-service-cli` is not provided as a image and can be removed
- I added some extra in the regex to avoid more than one white space
- Instead of using count, we can use the shell return code (`$?`), which is the standard.
- neat: multiline for better reading

rdimitrov (approved):
Looks great 👍

kairoaraujo (approved):

PR #143, opened 2022-12-09
feature: rstuf-worker for managing targets via SQL

kairoaraujo:
The motivation for this feature is the issue:
- Improve/review scalability for repository-service-tuf-worker

This implements a layer to store all the TUF Metadata `TargetFiles`.
The target files are stored in a SQL Database.
It gives more flexibility and performance to handle high traffic of
adding or removing `TargetFiles` to the metadata.

This change implements:

**Add the SQL Infrastructure**

- It adds the requirements (Python Requirments and Docker requirements)
- Implement the Docker development environment
- Adds the SQLAlchemy as the interface to the SQL, which enables
  multiple SQL Server support (we use for development PostgreSQL)
- Includes the Alambic for database migrations, including the initial
  migrations.
  -  It creates the folder `alembic/` and `alembic.ini`
  - The alembic is added in the exclude for the test as it is generated
    automatically.
  - Added to the `Makefile` functionality for db-migrations
  - A reference was added in `README.rst`
  - A new `entrypoint-dev.sh` was create to manage the migrations
    before starting the service
  - The `entrypoint-dev.sh` was changed to support migrations in
    deployment

**Add RSTUF database model**

Adds the RSTUF Worker database model
This implements the RSTUF Targets database table model in the RSTUF
Worker.

All the SQL Database model is added in the
`repository_service_tuf_worker/models`.

It implements:

- Schemas (data structures)
- SQL Models (using SQLAlchemy)
- CRUD

**Code functionality**

- The `repository.MetadataRepository` implements the `self._db`
  connected to the RSTUF database model from the `settings_refresh`.
- The `TargetFiles` are not added or removed directly manipulating
  the Metadata from the Storage Backend, but everything related to
  `TargetFiles` is managed in the DB SQL. Removing all the locks
  and growing the performance.
- The `update_targets` is responsible for managing the Metadata in
  the Storage Backend, using data from the DB SQL
- All unit tests
- All the design documentation


docker-compose.yml
```diff
@@ -56,24 +70,25 @@ services:
...
-      - RSTUF_LOCAL_KEYVAULT_PATH="/var/opt/repository-service-tuf/keystorage"
-      - RSTUF_BROKER_SERVER="amqp://guest:guest@rabbitmq:5672"
-      - RSTUF_REDIS_SERVER="redis://redis"
-      - RSTUF_WORKER_ID="dev"
+      - RSTUF_STORAGE_BACKEND=LocalStorage
+      - RSTUF_KEYVAULT_BACKEND=LocalKeyVault
+      - RSTUF_LOCAL_STORAGE_BACKEND_PATH=/var/opt/repository-service-tuf/storage
+      - RSTUF_LOCAL_KEYVAULT_PATH=/var/opt/repository-service-tuf/keystorage
+      - RSTUF_BROKER_SERVER=amqp://guest:guest@rabbitmq:5672
+      - RSTUF_SQL_SERVER=postgresql://postgres:secret@postgres:5432
```
rdimitrov, line 81: Shouldn't this be`5433`? Because of the port number update we do for the `postgres` service above?

docker-compose.yml
```diff
@@ -56,24 +70,25 @@ services:
...
+      - RSTUF_REDIS_SERVER=redis://redis
+      - RSTUF_WORKER_ID=dev1
     volumes:
       - ./:/opt/repository-service-tuf-worker:z
       - repository-service-tuf-storage:/var/opt/repository-service-tuf/storage
       - repository-service-tuf-keystorage:/var/opt/repository-service-tuf/keystorage
-    ports:
-      - "6900:6900"
```
rdimitrov, line 74: Is this correct that we do not expose anything from the `worker` anymore?

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
+                role.signed.targets = {
+                    target[0]: TargetFile.from_dict(target[1], target[0])
+                    for target in targets_crud.read_all_add_by_rolename(
+                        self._db, rolename
+                    )
+                }
+
+                # update expiry, bump version and persist to the storage
+                self._bump_expiry(role, BINS)
+                self._bump_version(role)
```
rdimitrov, line 466: Does that mean that the version is updated for each delegated role that has targets updated? For example, if we update several targets spread across 5 delegated roles, the version bump will be +5?

rdimitrov (approved):
Even though it's a relatively big code update most of the changes are relatively easy to follow 👍 Left a few comments but overall it looks like a nice performance optimization 👍

repository_service_tuf_worker/repository.py
kairoaraujo: No, the bump will always be by +1.
Not on line 441, we start the loop by rolename. On line 455 I load the metadata from the role, add all targets, and bump it to the current version +1

PR #145, opened 2022-12-13
feature: rstuf-worker for managing targets via SQL (#143)

kairoaraujo:
The motivation for this feature is the issue:
- Improve/review scalability for repository-service-tuf-worker

This implements a layer to store all the TUF Metadata `TargetFiles`. The target files are stored in a SQL Database.
It gives more flexibility and performance to handle high traffic of adding or removing `TargetFiles` to the metadata.

This change implements:

**Add the SQL Infrastructure**

- It adds the requirements (Python Requirments and Docker requirements)
- Implement the Docker development environment
- Adds the SQLAlchemy as the interface to the SQL, which enables multiple SQL Server support (we use for development PostgreSQL)
- Includes the Alambic for database migrations, including the initial migrations.
  -  It creates the folder `alembic/` and `alembic.ini`
  - The alembic is added in the exclude for the test as it is generated automatically.
  - Added to the `Makefile` functionality for db-migrations
  - A reference was added in `README.rst`
  - A new `entrypoint-dev.sh` was create to manage the migrations before starting the service
  - The `entrypoint-dev.sh` was changed to support migrations in deployment

**Add RSTUF database model**

Adds the RSTUF Worker database model
This implements the RSTUF Targets database table model in the RSTUF Worker.

All the SQL Database model is added in the
`repository_service_tuf_worker/models`.

It implements:

- Schemas (data structures)
- SQL Models (using SQLAlchemy)
- CRUD

**Code functionality**

- The `repository.MetadataRepository` implements the `self._db` connected to the RSTUF database model from the `settings_refresh`.
- The `TargetFiles` are not added or removed directly manipulating the Metadata from the Storage Backend, but everything related to `TargetFiles` is managed in the DB SQL. Removing all the locks and growing the performance.
- The `update_targets` is responsible for managing the Metadata in the Storage Backend, using data from the DB SQL
- All unit tests
- All the design documentation


kairoaraujo:
Sharing some additional information for the reviewers.
- FT test to this feature branch with our current 'main' branches for CLI and API.
https://github.com/vmware/repository-service-tuf/actions/runs/3684641079/jobs/6234613758

- FT + New FT performance to this features branch with our current 'main' branches for CLI and API
https://github.com/vmware/repository-service-tuf/actions/runs/3686525466/jobs/6238870323

This feature also requires the following:
- https://github.com/vmware/repository-service-tuf-api/pull/203# **_[it can be merged before]_**
- https://github.com/vmware/repository-service-tuf/pull/171 _**[it can be merged after]**_

rdimitrov (dismissed):

repository_service_tuf_worker/models/targets/crud.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        .filter(
+            models.RSTUFTargets.rolename == rolename,
+            models.RSTUFTargets.action == schemas.TargetAction.ADD,
+        )
+        .all()
+    )
+
+
+def update(
```
MVrachev, line 100: This will overwrite the old targets or add new RSTRUFTargets with an "ADD" action and keep the old one?

repository_service_tuf_worker/repository.py
```diff
@@ -251,7 +263,11 @@ def _bump_version(self, role: Metadata) -> None:
         """Bumps metadata version by 1."""
         role.signed.version += 1
 
-    def _update_timestamp(self, snapshot_version: int) -> Metadata[Timestamp]:
+    def _update_timestamp(
+        self,
+        snapshot_version: int,
+        db_targets: Optional[List[targets_models.RSTUFTargets]] = None,
```
MVrachev, line 269: Can you add a docstring about `db_targets` as the name doesn't show what are you supposed to pass there.

repository_service_tuf_worker/repository.py
```diff
@@ -287,87 +308,89 @@ def _update_snapshot(
...
-                if snapshot_meta_file.version == target_meta[1]:
-                    logging.debug(f"Found published meta {target_meta}")
-                    targets_meta.remove(target_meta)
-
-            if len(targets_meta) > 0:
-                _update_state(targets_meta)
+            completed_roles: List[str] = []
+            for role_name, targets in bin_targets.items():
+                for target in targets:
+                    self._db.refresh(target)
```
MVrachev, line 349: What exactly does this do?
Does it update the database with this new `target` information or what?

repository_service_tuf_worker/repository.py
```diff
@@ -287,87 +308,89 @@ def _update_snapshot(
...
                 meta={
-                    "unpublished_roles": unpublised_roles,
+                    "published_roles": completed_roles,
+                    "roles_to_publish": f"{list(bin_targets.keys())}",
                     "status": "Publishing",
                     "last_update": datetime.now(),
                 },
             )
 
         while True:
```
MVrachev, line 345: Do you need the `while True` loop given that if the `completed_roles` list is different than the `bin_targets` list you will call the `_update_state` function anyway?

repository_service_tuf_worker/repository.py
```diff
@@ -287,87 +308,89 @@ def _update_snapshot(
...
+    def _send_publish_targets_task(self):  # pragma: no cover
+        """
+        Send a new task to the `rstuf_internals` queue to publish targets.
+        """
+        # it is imported in the call to avoid a circular import
+        from app import repository_service_tuf_worker
+
+        # TODO: all tasks has the same id `publish_targets`. Should be unique?
+        # Should we check and avoid multiple tasks? Check that the function
+        # `publish_target` has a lock to avoice race condition.
```
MVrachev, line 372: - Nit:

```suggestion
        # `publish_target` has a lock to avoid race conditions.
```

repository_service_tuf_worker/repository.py
```diff
@@ -287,87 +308,89 @@ def _update_snapshot(
...
+            task_id="publish_targets",
+            queue="rstuf_internals",
+            acks_late=True,
+        )
 
     def bootstrap(
         self,
         payload: Dict[str, Dict[str, Any]],
-        update_state: Optional[str] = None,
+        update_state: Optional[str] = None,  # It is required (see: app)
```
MVrachev, line 386: Nit:
```suggestion
        update_state: Optional[str] = None,  # It is required (see: app.py)
```

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
+                db_targets = targets_crud.read_unpublished_by_rolename(
+                    db=self._db, rolename=rolename
+                )
+                logging.debug(f"{rolename}: New targets #: {len(db_targets)}")
+                db_published_targets += db_targets
+
+                # load the delegated targets role, clean the targets and add
+                # a new meta from the SQL DB.
+                # note: it might include targets from another parent task, it
+                # will speed the process of publishing new targets.
```
MVrachev, line 454: Nit:
```suggestion
                # will speed up the process of publishing new targets.
```

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
+            unpublished_roles = targets_crud.read_unpublished_rolenames(
+                self._db
+            )
 
-        Assignment is based on the hash prefix of the target file path. All
-        metadata is signed and persisted using the configured key and storage
-        services.
+            if len(unpublished_roles) == 0:
+                logging.debug("No new targets in delegated roles. Finishing")
+                return None
```
MVrachev, line 435: Maybe you can also release the redis lock before returning so that the lock is not taken for whole 5 seconds.

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
+
+            # initialize the new snapshot targets meta and published targets
+            # from DB SQL
+            new_snapshot_targets_meta: List[Tuple(str, int)] = []
```
MVrachev, line 439: I think you can shorten your variable name here with `new_snapshot_meta` as by definition the TUF spec snapshot meta contains information about all target metadata files.

repository_service_tuf_worker/repository.py
```diff
@@ -471,39 +557,33 @@ def remove_targets(
...
-                        )
-
-                    else:
-                        not_found_targets.append(path)
+                bin_targets[bins_name].append(db_target)
 
-        if len(targets_meta) > 0:
-            self._add_to_unpublished_metas(targets_meta)
-            self._publish_meta_state(targets_meta, update_state)
+        if len(deleted_targets) > 0:
```
MVrachev, line 584: I noticed you are keeping track of the targets for deletion in `deleted_targets` in `remove_targets`, but you don't keep track of the targets for addition in add_targets.
Why is that and what's the difference between `add_targets` and `remove_targets`?

MVrachev (changes requested):
Here we have a great amount of work.
Awesome @kairoaraujo!

I have multiple questions, but I guess you can expect that from me when there is such a big pr.

In addition to all of my notes/questions I will add that I don't see any changes in
`docs/source/_static/repository-service-tuf-worker-C1.png`, but I see it's added as a changed file. [🤷](https://emojipedia.org/person-shrugging/)

Dockerfile
```diff
@@ -11,15 +11,16 @@ ENV PYTHONDONTWRITEBYTECODE=1
 ADD requirements.txt /builder/requirements.txt
 
 WORKDIR /builder
-RUN apt-get update && apt-get install gcc -y
+RUN apt-get update && apt-get install gcc libpq-dev -y
+RUN pip install --upgrade pip && pip install --user -r requirements.txt
```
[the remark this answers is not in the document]
kairoaraujo: We have to do it, once on line 16. We remove the compiler (`gcc` and libraries) used by some packages installed by pip (some packages require `gcc`, such as `psycopg2`.

We install the required compilers and os (gcc, libpq-dev), install python libraries than clean up from the docker image the gcc and sub-dependencies. libpq-dev needs to stay because psycopg2 uses some linked libraries.

Dockerfile
```diff
@@ -11,15 +11,16 @@ ENV PYTHONDONTWRITEBYTECODE=1
 ADD requirements.txt /builder/requirements.txt
 
 WORKDIR /builder
-RUN apt-get update && apt-get install gcc -y
+RUN apt-get update && apt-get install gcc libpq-dev -y
```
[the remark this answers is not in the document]
kairoaraujo: `psycopg2` requires `libpq-dev`

Dockerfile
```diff
@@ -11,15 +11,16 @@ ENV PYTHONDONTWRITEBYTECODE=1
...
+RUN pip install --upgrade pip && pip install --user -r requirements.txt
 RUN apt-get remove gcc --purge -y \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean autoclean \
     && apt-get autoremove --yes
-RUN pip install --upgrade pip && pip install --user -r requirements.txt
 
 # Final image
 FROM base_os as pre-final
+RUN apt-get update && apt-get install libpq-dev -y && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
kairoaraujo: It is not long-used data. It helps to have a small RSTUF worker image.

docker-compose.yml
```diff
@@ -1,12 +1,13 @@
 version: "3.9"
 
 volumes:
-  repository-service-tuf-mq-data:
-  repository-service-tuf-api-data:
   repository-service-tuf-storage:
   repository-service-tuf-keystorage:
+  repository-service-tuf-api-data:
+  repository-service-tuf-mq-data:
```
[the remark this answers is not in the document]
kairoaraujo: The first two are related to workers, and the two others are related to the API Service.
I just did it as the docker-compose is for the worker to have it highlighted.

repository_service_tuf_worker/models/__init__.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: 2022 VMware Inc
+#
+# SPDX-License-Identifier: MIT
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from repository_service_tuf_worker.models.targets import (  # noqa
```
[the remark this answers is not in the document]
kairoaraujo: We import, but we don't use these imports here.
They are imported here to make the Python package simpler to use in the other files.
Example:
```python
from repository_service_tuf_worker.models import (
     rstuf_db,
     targets_crud,
     targets_models,
     targets_schema,
 )
```

repository_service_tuf_worker/models/targets/crud.py
```diff
@@ -0,0 +1,149 @@
...
+    return (
+        db.query(models.RSTUFTargets)
+        .filter(models.RSTUFTargets.path == path)
+        .first()
+    )
+
+
+def read_by_rolename(db: Session, rolename: str) -> List[models.RSTUFTargets]:
+    """
+    Read all Targets by the delegated role name.
```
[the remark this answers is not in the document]
kairoaraujo: Here I read all Targets in a specific role name.
You can have multiple target files in the same delegated role name.

repository_service_tuf_worker/models/targets/crud.py
kairoaraujo: It overwrites the existent RSTFUTargets with ADD. 
Even an existent RSTUFTarget that has action as `REMOVE`, if updated, must move to `ADD`.
A delete/remove action never operates in an existent target with the action 'REMOVE'. It is 'REMOVED' already.

repository_service_tuf_worker/repository.py
kairoaraujo: Not clear if I understood the question.
Yes, I need it. Otherwise, the task initiated in the `add_targets` or `remove_targets` will finish.
The task needs to keep running until it finishes.

repository_service_tuf_worker/repository.py
kairoaraujo: The SQLAlchemy DB Session updates the target object that another thread in the SQL might update.

repository_service_tuf_worker/repository.py
kairoaraujo: We use context (see line 427)
When it returns the lock is released.

repository_service_tuf_worker/repository.py
kairoaraujo: The difference is that ``add_targets`` does not constrain whether the target exists. Once the target data in the payload is valid, it is ok.

Removing targets, there is a constraint if the target exists or not.
A user can require to ``remove_targets`` and part to all does not exist.

In the task response, we tell the user the deleted/removed targets and the not found targets.

docker-compose.yml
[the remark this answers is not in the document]
MVrachev: I see. It's good to clarify those change somewhere and that's why I asked you.

repository_service_tuf_worker/models/__init__.py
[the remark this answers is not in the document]
MVrachev: If we don't use those packages in this module it's good to leave a comment about why we import them at all.

repository_service_tuf_worker/models/targets/crud.py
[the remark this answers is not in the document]
MVrachev: Oh, yes, my mistake.
I understood something else  that there could be two targets with the exact same name.

repository_service_tuf_worker/models/__init__.py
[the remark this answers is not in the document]
kairoaraujo: I think it is a quite usual standard in python nested packages. Some examples

- https://github.com/pallets/flask/blob/main/src/flask/__init__.py
- https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/__init__.py

repository_service_tuf_worker/models/targets/crud.py
MVrachev: > A delete/remove action never operates in an existent target with the action 'REMOVE'. It is 'REMOVED' already.

Sorry, but I am not sure I understood what you mean here.

repository_service_tuf_worker/repository.py
MVrachev: What I mean here is that without the `while True` look you will go inside the
`for role_name, targets in bin_targets.items():`  and add some roles in `completed_roles`.
Then at line 357 you will check `if sorted(completed_roles) != sorted(list(bin_targets)):` and if that's true, then you will call update_state again and redo the same thing.
Not sure what is the benefit `while True` brings here as you will traverse the same code again if you are not finished.

repository_service_tuf_worker/models/targets/crud.py
kairoaraujo: This CRUD is an Update and always overwrites an existent RSTUFTarget with "ADD".
The only case we operate on an existent RSTUFTarget is in `add_targets`

https://github.com/vmware/repository-service-tuf-worker/blob/e5e398ecd601f047b65175f542b09445d08fc905/repository_service_tuf_worker/repository.py#L569-L575

If an RSTUFTarget has an action `REMOVED` and `published` is False, we just reinforce it:

https://github.com/vmware/repository-service-tuf-worker/blob/38ea42cd64ea8b683b00e60da7f37b2f9d3dc806/repository_service_tuf_worker/repository.py#L574-L577

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
-            bin_target_groups: Dict[str, List[TargetFile]] = {}
-            for target in targets:
-                bins_name = self._get_path_succinct_role(target.get("path"))
-                if bins_name not in bin_target_groups:
-                    bin_target_groups[bins_name] = []
-
-                target_file = TargetFile.from_dict(
-                    target["info"], target["path"]
+        # Group target files by responsible 'bins' delegated roles.
+        # This will be used to by `publish_targets`
```
MVrachev, line 503: Nit:
```suggestion
        # This will be used by `publish_targets`
```

repository_service_tuf_worker/repository.py
MVrachev: Sorry, but I read line 427 `with self._redis.lock("publish_targets", timeout=5.0):` and I don't see where we actually say we use context.
Am I right to guess that at line `427` we lock the `publish_targets` which will be released by another function somewhere else?

repository_service_tuf_worker/repository.py
MVrachev: Ahaa, yes, I see what you mean.

MVrachev (commented):
Left a couple more questions.

repository_service_tuf_worker/repository.py
kairoaraujo: But I will run it only once, update_state once, return None and finish the task at line:

https://github.com/vmware/repository-service-tuf-worker/blob/38ea42cd64ea8b683b00e60da7f37b2f9d3dc806/repository_service_tuf_worker/repository.py#L542

or

https://github.com/vmware/repository-service-tuf-worker/blob/38ea42cd64ea8b683b00e60da7f37b2f9d3dc806/repository_service_tuf_worker/repository.py#L598

Before I finished all targets.

repository_service_tuf_worker/repository.py
kairoaraujo: We use the context `with self._redis.lock()`
Same as we use `with open()` for files. You never close the file. You will be using the context. When the block finishes, automatically close the file. The same applies to the lock. When the block ends, it releases the lock.

repository_service_tuf_worker/repository.py
kairoaraujo: It refreshes from the database, to read the latest committed data.

repository_service_tuf_worker/repository.py
MVrachev: Got it, thanks.

docs/source/devel/index.rst
```diff
@@ -94,66 +94,43 @@ We use the 'waiting time' to alternate between tasks.
...
-      }
-   @enduml
+The task will continue run, wait until all the targets are persisted to the
+Repository Metadata backend.
 
-Every minute, the routine task **Publish the new Hashed Bins Roles** runs and
-gets the names of the unpublished Hashed Bins Roles, looks in the Storage the
-latest version and runs steps 4 to 9, and flushes the ``unpublished_meta``.
+The **Publish the new Hashed Bins Roles** task (``publish_targets``) runs once
+per time to using locks [#f1]_ . It will  will do:
```
MVrachev, line 110: Could you add a sentence or two about how we synchronize the state of the database with the state of the JSON metadata files?

repository_service_tuf_worker/repository.py
MVrachev: I read the function again locally and it seems I read it incorrectly.
Thanks.

docs/source/devel/index.rst
kairoaraujo: @MVrachev  Added the mention.

docs/source/devel/index.rst
```diff
@@ -50,19 +50,20 @@ The repository Worker has two maintenance tasks:
...
-**Publish the new Hashed Bins Roles** (``publish_targets_meta``) is part of the
+**Publish the new Hashed Bins Target Roles** (``publish_targets``) is part of the
 solution for the :ref:`Repository Worker scalability, Issue 17
 <devel/known_issues:(Solved) Scalability>`.
 
 To understand more, every time the API sends a task to add a new target, the
-Hashed Bins Roles must be changed to add the new target(s), as the Snapshot and
-Timestamp.
+Hashed Bins Roles must be changed to add the new target(s), followed by a new
+Snapshot and Timestamp versions.
```
MVrachev, line 66: Even though timestamp, snapshot and hashed bins are important for us I don't see why they are with upper case letters.
I have the same opinions for `Repository Worker` and `Repository Metadata backend`.

What do you think?

docs/source/devel/index.rst
```diff
@@ -94,66 +95,48 @@ We use the 'waiting time' to alternate between tasks.
...
 
-It means the multiple Repository Workers can write multiple Hashed Bins Roles
-simultaneously from various tasks.
+The **Publish the new Hashed Bins Target Roles** is a synchronization between
+the SQL Database and the Hashed Bins Target Roles in the Backend Storage (i.e.
+JSON files in the filestytem)
 
-When Step 3 finishes, the Repository Worker will lock  ``unpublished_meta`` and
-add the Hashed Bins Role in the list if it does not exist.
+When a task finishes, it send a task the **Publish the new Hashed Bins Target
```
MVrachev, line 107:
```suggestion
When a task finishes, it sends a task the **Publish the new Hashed Bins Target
```

docs/source/devel/index.rst
```diff
@@ -94,66 +95,48 @@ We use the 'waiting time' to alternate between tasks.
...
-               :Add Hashed Bins name;
-            endif
-            :Unlock unpublished_meta;
-            :Unlock Hashed Bins Roles;
-         repeat
-         repeat while (all Hashed Bins are Published) is (Waiting)
-         stop
-      }
-   @enduml
+The task will continue run, wait until all the targets are persisted to the
```
MVrachev, line 113:
```suggestion
The task will continue running and waiting until all the targets are persisted to the
```

docs/source/devel/index.rst
```diff
@@ -94,66 +95,48 @@ We use the 'waiting time' to alternate between tasks.
...
+Repository Metadata backend.
 
-Every minute, the routine task **Publish the new Hashed Bins Roles** runs and
-gets the names of the unpublished Hashed Bins Roles, looks in the Storage the
-latest version and runs steps 4 to 9, and flushes the ``unpublished_meta``.
+The **Publish the new Hashed Bins Target Roles** runs once per time to using
```
MVrachev, line 116:
```suggestion
The **Publish the new Hashed Bins Target Roles** task runs once per time to using
```

MVrachev (commented):
Great improvements @kairoaraujo!
Given the size of the changes, it's inevitable for me to have more questions and comments.

docs/source/devel/index.rst
kairoaraujo: I use it to highlight the role name (not to confuse it with the generic word snapshot timestamp) and the component name (Repository Worker).
We do it across many docs. My suggestion is if we want to normalize it do a specific issue and change across all doc.

kairoaraujo:
@MVrachev changes applied.

repository_service_tuf_worker/repository.py
```diff
@@ -398,61 +417,128 @@ def bootstrap(
...
+        bin_targets: Dict[str, List[targets_models.RSTUFTargets]] = {}
+        for target in targets:
+            bins_name = self._get_path_succinct_role(target.get("path"))
```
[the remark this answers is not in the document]
kairoaraujo: As it is mandatory, doesn't it matter if parsing with `.get("path")` or `["path"]`?

repository_service_tuf_worker/repository.py
```diff
@@ -92,6 +99,7 @@ def __init__(self):
         self._settings = repository_settings
         self._storage_backend = self.refresh_settings().STORAGE
         self._key_storage_backend = self.refresh_settings().KEYVAULT
+        self._db = self.refresh_settings().SQL
```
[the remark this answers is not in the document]
kairoaraujo: That way, we refresh every time we call the variable `self._`.
It is important in a distributed environment with multiple workers that it is refreshed in run-time if someone updates the worker settings.

repository_service_tuf_worker/repository.py
```diff
@@ -398,112 +425,172 @@ def bootstrap(
...
+            # note: the `db_published_targes` contains the targets that
+            # needs to updated in SQL DB as 'published' and it will be done
+            # by the `_update_timestamp`
+            self._update_timestamp(
+                self._update_snapshot(new_snapshot_meta),
+                db_published_targets,
+            )
+
+    def add_targets(
+        self, payload: Dict[str, Any], update_state: str
```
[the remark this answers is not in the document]
kairoaraujo: I added the correct typing.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -462,40 +476,180 @@ def test_bootstrap_missing_metadata(self):
...
+            pretend.call(
+                "fake_md_snapshot",
+                [
+                    fake_db_targets,
+                    fake_db_targets,
+                    fake_db_targets,
+                    fake_db_targets,
+                ],
+            )
+        ]
```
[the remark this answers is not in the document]
kairoaraujo: This is because we record and assert each unit of the function. But we can discuss it separated.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -525,23 +681,86 @@ def mocked_lock(lock):
...
-            pretend.call([("bin-e", 41)])
+        assert test_repo._send_publish_targets_task.calls == [
+            pretend.call("fake_task_id_xyz")
         ]
-        assert test_repo._publish_meta_state.calls == [
-            pretend.call([("bin-e", 41)], fake_ue)
+        assert test_repo._update_task.calls == [
+            pretend.call({"bin-e": [fake_db_target]}, fake_ue)
         ]
         assert fake_datetime.now.calls == [pretend.call()]
```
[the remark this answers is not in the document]
kairoaraujo: It looks the same, but there is a slight difference.
`test_add_targets` in the DB mock return `none` that will call the CRUD `create`
`test_add_targets_exists in the DB mock, it returns already the `fake_db_target` that will call the CRUD `update`
But it was not clear. I made it correct now.

repository_service_tuf_worker/repository.py
```diff
@@ -398,112 +427,172 @@ def bootstrap(
...
+                self._db
+            )
+
+            if len(unpublished_roles) == 0:
+                logging.debug("No new targets in delegated roles. Finishing")
+                return None
+
+            # initialize the new snapshot targets meta and published targets
+            # from DB SQL
+            new_snapshot_meta: List[Tuple(str, int)] = []
```
MVrachev, line 449: I suggested it before in https://github.com/vmware/repository-service-tuf-worker/pull/145#discussion_r1049654442 and you resolved it, but it was somehow lost.

repository_service_tuf_worker/repository.py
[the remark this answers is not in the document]
MVrachev: Ahaa, you mean if someone updates the worker settings while you were updating all three of them at once?

repository_service_tuf_worker/repository.py
```diff
@@ -398,112 +427,172 @@ def bootstrap(
...
+            # needs to updated in SQL DB as 'published' and it will be done
+            # by the `_update_timestamp`
+            self._update_timestamp(
+                self._update_snapshot(new_snapshot_meta),
+                db_published_targets,
+            )
+
+    def add_targets(
+        self, payload: Dict[str, Any], update_state: Task.update_state
+    ) -> Dict[str, Any]:
```
MVrachev, line 493: It seems the docstring you added for `payload` and `update_state` has been removed somewhere.
Could you add it again?
We discussed this in https://github.com/vmware/repository-service-tuf-worker/pull/145#discussion_r1056417965,

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
```diff
@@ -0,0 +1,281 @@
...
+        )
+        mocked_db = pretend.stub(
+            query=pretend.call_recorder(lambda *a: mocked_db_filter)
+        )
+
+        test_result = crud.read_unpublished_rolenames(mocked_db)
+
+        assert test_result == [(False, "bin-e"), (False, "bin-3")]
+        assert mocked_db.query.calls == [pretend.call(False, "all-bins")]
+        assert mocked_db_filter.filter.calls == [pretend.call(True)]
```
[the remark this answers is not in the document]
MVrachev: I am sorry, but I didn't see your response to this question.

tests/unit/tuf_repository_service_worker/test_repository.py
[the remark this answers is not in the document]
MVrachev: Yes, just wanted to raise the topic that we need to think about how can do we do that better.
No need to fix this in this big pr.

docs/source/devel/index.rst
MVrachev: Okay, let's leave it like that for now.

MVrachev (changes requested):
I noticed some of the comments I made that were marked as resolved weren't actually resolved.
As I mention in my comments it's likely due to a mistake during a rebase.

repository_service_tuf_worker/models/targets/schemas.py
```diff
@@ -0,0 +1,22 @@
...
+    ADD = "ADD"
+    REMOVE = "REMOVE"
+
+
+class TargetsCreate(BaseModel):
+    path: str
+    info: Dict[str, Any]
+    rolename: str
+    published: bool
+    action: Literal[TargetAction.ADD, TargetAction.REMOVE]
```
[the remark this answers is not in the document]
MVrachev: Sorry, but I was curious about that one and that's why I will mark it as unresolved. See https://github.com/vmware/repository-service-tuf-worker/pull/145#discussion_r1056389410.

repository_service_tuf_worker/repository.py
kairoaraujo: You've suggested shortening the name from  `new_snapshot_targets_meta` to `new_snapshot_meta`. Not sure what I'm missign.

repository_service_tuf_worker/repository.py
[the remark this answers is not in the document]
kairoaraujo: Yes, that is correct. If someone changes one worker's settings, the settings are stored in Redis. All workers will have it fresh in the next call.

repository_service_tuf_worker/repository.py
kairoaraujo: As you asked to add docstring because of the wrong typing for `update_state`,  I fixed the correct typing, which makes more clear.

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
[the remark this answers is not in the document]
kairoaraujo: OK, I left it to discuss in a call, but writing here.

We are testing `test_read_unpublished_rolenames`.
It uses the SQLAchemy and we use the db session created by it that has the built-ins `query`, `filter` etc.

The [pretend.](https://github.com/alex/pretend)`call_recorder` recorders the call parameters used by the function, not the return. That way, you can track each unit of your code and see if it is calling corrected, even if it is mocked.

The `mocked_db_filter.filter.calls` will have a "pretend call" with `True` because we use pretend for mocking it. We are not testing what SQLAlchemy is using here.
We could even get ride-off `mock_db_dilter.calls` here, to be honest. The important is the query call.

repository_service_tuf_worker/repository.py
```diff
@@ -287,87 +315,92 @@ def _update_snapshot(
...
+    def _send_publish_targets_task(self, task_id: str):  # pragma: no cover
+        """
+        Send a new task to the `rstuf_internals` queue to publish targets.
+        """
+        # it is imported in the call to avoid a circular import
+        from app import repository_service_tuf_worker
+
+        # TODO: all tasks has the same id `publish_targets`. Should be unique?
+        # Should we check and avoid multiple tasks? Check that the function
+        # `publish_target` has a lock to avoice race conditions.
```
MVrachev, line 380:
```suggestion
        # `publish_target` has a lock to avoid race conditions.
```

repository_service_tuf_worker/repository.py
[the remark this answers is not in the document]
MVrachev: We can create an issue about this by clarifying what we expect the `payload` dictionary to be and discussing in general how we can explain that `payload` is the "contract" between the RSTUF API and Worker.

repository_service_tuf_worker/repository.py
MVrachev: You are right, my mistake.

MVrachev (approved):
I like what you did here.
I think we managed to improve it even more after the review iterations.

There is only one issue I see that we can create about refactoring the names of the variables in `test_crud.py`.

PR #158, opened 2023-01-05
Synchronize changes in Makefile with API

MVrachev:
Synchronize changes in Makefile with API

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,11 @@ stop:
 clean:
 	$(MAKE) stop
 	docker-compose rm --force
+	rm -rf ./metadata/*
+	rm -rf ./keys/*
+	rm -rf ./database/*.sqlite
 	rm -rf ./data
+	rm -rf ./data_test
 
 purge:
 	$(MAKE) clean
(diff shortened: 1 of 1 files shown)


kairoaraujo (approved):

Makefile
```diff
@@ -18,7 +18,11 @@ stop:
 clean:
 	$(MAKE) stop
 	docker-compose rm --force
+	rm -rf ./metadata/*
+	rm -rf ./keys/*
+	rm -rf ./database/*.sqlite
```
kairoaraujo, line 23: These data are not generated using the API if you run `make run`.
I am not against adding those, but not sure if it is necessary.

kairoaraujo (approved):

Makefile
MVrachev: Then should we remove those lines from the API as well: 
https://github.com/vmware/repository-service-tuf-api/blob/51440aa3315b915c0e499ee91aba3a2b19a68ea0/Makefile#L38

kairoaraujo (approved):

PR #159, opened 2023-01-05
Fix sslib new versions incompatibility bug

MVrachev:
The incompatibility with newer securesystemslib versions was caused because of a new breaking change introduced in:
https://github.com/secure-systems-lab/securesystemslib/pull/231

The changes were tested locally and the bootstrap works.

It seems that `Pipenv.lock` wasn't updated for a while as many changes not related were added.
I ran `pipenv lock` and on the time of submitting this pr on the latest main 248 lines of insertions and 200 lines of deletion are introduced in `Pipenv.lock`.

I added unit tests testing the new `restrict` argument to the `LocalStorage.put()` call which is used by default.
Also, updated the old tests to be consistent with the newer code.

Closes https://github.com/vmware/repository-service-tuf-worker/issues/118


MVrachev:
There seems to be a lint failure. I will work to fix it today.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 -i https://pypi.org/simple
-alembic==1.8.1
```
MVrachev, line 2: When I updated the `securesystemslib` I noticed that `tox -e requirements` failed suddenly.
I thought it was somehow connected to the newer securesytemslib versions but in their requirements files, I don't see those dependencies...

kairoaraujo: I merged #156 
If you rebase it now, the changes only related to your PR will be clear.

MVrachev:
Rebased and it's ready for review.

kairoaraujo:
Can you please rebase again?

MVrachev:
@kairoaraujo I rebased.

kairoaraujo (dismissed):
LGTM
nit: You can update the dependencies, sslib 0.26.0 was released, or let our workflow update after the merge.
```shell
$ pipenv update
$ make requirements
```

```diff
$  git diff
diff --git a/Pipfile.lock b/Pipfile.lock
index 3b5333c..4855f21 100644
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -567,11 +567,11 @@
         },
         "securesystemslib": {
             "hashes": [
-                "sha256:04bc11593edd68405939d3dfc318080bfb31f1ebb5d81c7911914b42dfd4bf2f",
-                "sha256:10d5a066e70cb87704c9bf2cef1ef6d8a06fab5ef7602dd59c26d06251317a11"
+                "sha256:41c7b25c52dc0bafe774413b5738bbf4431f094e72a091e83d9921901972ae4c",
+                "sha256:a8fa49831d6a7e48f81050984ddfac3713af0c326f558727113533edb5ca8eac"
             ],
             "index": "pypi",
-            "version": "==0.25.0"
+            "version": "==0.26.0"
         },
         "setuptools": {
             "hashes": [
@@ -1551,11 +1551,11 @@
         },
         "securesystemslib": {
             "hashes": [
-                "sha256:04bc11593edd68405939d3dfc318080bfb31f1ebb5d81c7911914b42dfd4bf2f",
-                "sha256:10d5a066e70cb87704c9bf2cef1ef6d8a06fab5ef7602dd59c26d06251317a11"
+                "sha256:41c7b25c52dc0bafe774413b5738bbf4431f094e72a091e83d9921901972ae4c",
+                "sha256:a8fa49831d6a7e48f81050984ddfac3713af0c326f558727113533edb5ca8eac"
             ],
             "index": "pypi",
-            "version": "==0.25.0"
+            "version": "==0.26.0"
         },
         "setuptools": {
             "hashes": [
diff --git a/requirements-dev.txt b/requirements-dev.txt
index 4c49be2..c969b33 100644
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -59,7 +59,7 @@ pytz==2022.7.1
 pyyaml==6.0 ; python_version >= '3.6'
 redis==4.4.2
 requests==2.28.2 ; python_version >= '3.7' and python_version < '4'
-securesystemslib==0.25.0
+securesystemslib==0.26.0
 setuptools==66.0.0 ; python_version >= '3.7'
 six==1.16.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
 snowballstemmer==2.2.0
diff --git a/requirements.txt b/requirements.txt
index f7b9136..a7a7973 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,7 +27,7 @@ pynacl==1.5.0
 pytz==2022.7.1
 redis==4.4.2
 requests==2.28.2 ; python_version >= '3.7' and python_version < '4'
-securesystemslib==0.25.0
+securesystemslib==0.26.0
 setuptools==66.0.0 ; python_version >= '3.7'
 six==1.16.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
 sqlalchemy==1.4.46
 ```

**Comment by MVrachev** [comment]
I removed the need for `StorageBackendInterface`, updated `LocalStorage.put` and updated the tests accordingly.
Ready for another review.

**Comment by MVrachev** [comment]
Rebased because of merge conflicts.

File: repository_service_tuf_worker/repository.py [hunk 2]
```diff
@@ -244,7 +245,12 @@ def _persist(self, role: Metadata, role_name: str) -> str:
             if filename[0].isdigit() is False:
                 filename = f"{role.signed.version}.{filename}"
 
-        role.to_file(filename, JSONSerializer(), self._storage_backend)
+        bytes_data = role.to_bytes(JSONSerializer())
+
+        with tempfile.TemporaryFile() as temp_file:
+            temp_file.write(bytes_data)
+
+            self._storage_backend.put(temp_file, filename)
```
kairoaraujo, line 253: Can you explain why we need a temporary file to send to backend storage?

MVrachev: Honestly, I just followed what python-tuf did with their `Metadata.to_file()` function: 
https://github.com/theupdateframework/python-tuf/blob/cc48e9d5f158022405349fe9854d0ae67c11ae53/tuf/api/metadata.py#L337-L339
The reason I can think of why `python-tuf` used that from the start is that they decided to use `tempFile` to make sure that if there is an error with a file name already existing they won't overwrite the old file.

MVrachev: We can add such a check and remove the `tempFile` usage.

MVrachev: Together with @kairoaraujo we discussed that there is no sense in writing the file inside the `_persist` method and instead we can move it into the `LocalStorage` implementation. 
Also, we decided to remove the usage of the `temp` module.

MVrachev:
Ready for another review @kairoaraujo.

kairoaraujo (approved):

Issue #157, opened 2023-01-05
I like what you did here.


kairoaraujo:
I like what you did here.
I think we managed to improve it even more after the review iterations.

There is only one issue I see that we can create about refactoring the names of the variables in `test_crud.py`.

_Originally posted by @MVrachev in https://github.com/vmware/repository-service-tuf-worker/pull/145#pullrequestreview-1237551739_

MVrachev:
I took the liberty to rename the issue to something more meaningful than `I like what you did here`  😀.
Also I removed one irrelevant to the issue sentence connected to the source pr.

PR #161, opened 2023-01-09
test: refactored test_crud variables name

kairoaraujo:
Refactored test_crud.py with more clear variables name.

Mocking the DB requires multiple sub-functions to be also mocked. This commit uses a more readable naming for these mocks.

Closes: #157 

### DIFF
--- a/tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
+++ b/tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
@@ -38,40 +38,40 @@ class TestTargetsCrud:
             published=False, rolename="all-bins"
         )
 
-        mocked_distinct = pretend.stub(
+        mocked_all = pretend.stub(
             all=pretend.call_recorder(
                 lambda: [(False, "bin-e"), (False, "bin-3")]
             )
         )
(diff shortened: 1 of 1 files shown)


MVrachev (commented):
Yea, probably better, but what if we have multiple stacked function calls?
Do we add all of them in the variable names?
For example `mocked_db_filder_distinct_group`?
Which one is the important one? The last word in this variable?

kairoaraujo:
> Yea, probably better, but what if we have multiple stacked function calls?
> Do we add all of them in the variable names?
> For example `mocked_db_filder_distinct_group`?
> Which one is the important one? The last word in this variable?

I think it is valid for readability, as it is a test, and we don't interact with this code (calling functions, etc).
I noticed only the initial part was committed. I adding all changes.

MVrachev (commented):
Why not just use the names of the functions of importance like:
`mocked_db`, `mocked_filter`, `mocked_distinct`.
What I think is better is to rename the current variables like that:
- `mocked_distinct` -> `mocked_all`
- `mocked_db_filter_distinct` -> `mocked_distinct`
- `mocked_db_filter` -> `mocked_filter`
- `mocked_db` will remain the same

It's seems easier for me to follow the functions.

PR #181, opened 2023-01-24
Add LocalKeyVault configuration settings for the online key

MVrachev:
Add configuration settings in LocalKeyVault for the online key that
will be used for "timestamp", "snapshot", "targets"  and "bin-n" roles.
That will allow us to load the online key instead of generating it
which is in line with what we want when we support more key vaults.

It's planned that if the user doesn't provide a path to the online key
as recommended we will generate an online key for him.

Additionally, add missing type annotations for the interfaces.

Close https://github.com/vmware/repository-service-tuf-worker/issues/169

### DIFF
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -106,10 +106,13 @@ Available types:
   - Optional variables:
     - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_NAME``
       - file name of the online key
+      - Default: `online.key`
     - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_PASSWORD``
       - password used to load the online key
     - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_TYPE``
-      - cryptographic type of the online key, example: `ed25519`, `rsa`
+      - cryptographic type of the online key, example: `ed25519`.
+      - Default: `ed25519`
+      - [Note: At the moment RSTUF Worker supports only `ed25519`
 
 
 #### (Optional) `DATA_DIR`
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -103,6 +103,13 @@ Available types:
 * LocalKeyVault (local file system)
   - Requires variable ``RSTUF_LOCAL_KEYVAULT_PATH``
     - Define the directory where the data will be saved, example: `keyvault`
+  - Optional variables:
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_NAME``
+      - file name of the online key
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_PASSWORD``
+      - password used to load the online key
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_TYPE``
+      - cryptographic type of the online key, example: `ed25519`, `rsa`
 
 
 #### (Optional) `DATA_DIR`
--- a/repository_service_tuf_worker/services/keyvault/local.py
+++ b/repository_service_tuf_worker/services/keyvault/local.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 import os
-from typing import Any, Dict, List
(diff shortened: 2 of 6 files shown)


repository_service_tuf_worker/services/keyvault/local.py
```diff
@@ -20,31 +20,63 @@ class KeyVaultError(Exception):
 class LocalKeyVault(IKeyVault):
     """Local KeyVault type"""
 
-    def __init__(self, path: str):
+    def __init__(
+        self,
+        path: str,
+        online_key_name: Optional[str] = "online.key",
+        online_key_pass: Optional[str] = None,
+        online_key_type: Optional[str] = "ed25519",
```
kairoaraujo, line 28: I was thinking now about this configuration, and to start, we should limit the support and extend it by demand.
We should start supporting only two key cryptography types. My suggestion 'ed25519' and 'RSA'
Otherwise, in the [Task: Remove LocalKeyVault functionality to write keys in the local storage](https://github.com/vmware/repository-service-tuf-worker/issues/170) we will add a lot of algorithms.
I suggest we add it as an `Optional` `enum` list with the default as `ed25519`

repository_service_tuf_worker/services/keyvault/local.py
```diff
@@ -20,31 +20,63 @@ class KeyVaultError(Exception):
...
+    ):
+        """A configuration class for RSTUF Worker services.
+
+        Args:
+            path: directory of the key vault.
+            online_key_name: file name of the online key.
+            online_key_pass: password to load the online key.
```
kairoaraujo, line 35: Document the supported list from the enum. (see my comment above)

kairoaraujo (changes requested):
Hi @MVrachev, you did good work here 🎉 
I left some comments in the code, and here are two suggestions as well
- Updated [Docker container instructions](ttps://github.com/vmware/repository-service-tuf-worker/blame/main/docs/source/guide/Docker_README.md#L103)
- Should this PR go to the `feature/role-simplification`?

repository_service_tuf_worker/services/keyvault/local.py
MVrachev: This was a topic of a long discussion.
We decided to validate the type and algorithm of the keys during the bootstrap ceremony and later when implemented during the rotation of the key as well.
We decided this to be a separate issue: https://github.com/vmware/repository-service-tuf-worker/issues/202

repository_service_tuf_worker/services/keyvault/local.py
MVrachev: See my comment above about the supported algorithms.

kairoaraujo (changes requested):
Can you please update the `Docker_README` (`docs/source/guide/Docker_README.md`)?
We need to add the new environment variables added in the change.
`

docs/source/guide/Docker_README.md
```diff
@@ -103,6 +103,13 @@ Available types:
 * LocalKeyVault (local file system)
   - Requires variable ``RSTUF_LOCAL_KEYVAULT_PATH``
     - Define the directory where the data will be saved, example: `keyvault`
+  - Optional variables:
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_NAME``
+      - file name of the online key
```
kairoaraujo, line 108:
```suggestion
      - file name of the online key. Default: `online.key`
```
In case the user decides to rename the key to the default value and not use a specific one.

docs/source/guide/Docker_README.md
```diff
@@ -103,6 +103,13 @@ Available types:
...
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_PASSWORD``
+      - password used to load the online key
+    - ``RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_TYPE``
+      - cryptographic type of the online key, example: `ed25519`, `rsa`
```
kairoaraujo, line 112:
```suggestion
      - cryptographic type of the online key. Default: `ed25519` [Note: At the moment RSTUF Worker supports only `ed25519`
```

kairoaraujo (approved):

PR #186, opened 2023-01-26
chore: improve performance updating Targets in DB

### LINKED ISSUE #176: Task: Improve database update for published targets
### What is the task about?

- This issue was originated by https://github.com/vmware/repository-service-tuf/issues/188

> [must] Improve the [database update for published target files](https://github.com/vmware/repository-service-tuf-worker/blob/7225eff9a487c0dccd0e771a39403ea1fc96d8d9/repository_service_tuf_worker/repository.py#L290). We do it one by one instead using a query/update in [CRUD](https://github.com/vmware/repository-service-tuf-worker/blob/7225eff9a487c0dccd0e771a39403ea1fc96d8d9/repository_service_tuf_worker/models/targets/crud.py#L121-L133). Also the place where we update the target from the database.

Improve the [database update for published targets files](https://github.com/vmware/repository-service-tuf-worker/blob/7225eff9a487c0dccd0e771a39403ea1fc96d8d9/repository_service_tuf_worker/repository.py#L290). 

We update all published hash bin delegated roles one by one instead of using a query/update in [CRUD](https://github.com/vmware/repository-service-tuf-worker/blob/7225eff9a487c0dccd0e771a39403ea1fc96d8d9/repository_service_tuf_worker/models/targets/crud.py#L121-L133). 

We update the database in the last action when we update the timestamp. 

### Parent feature

- https://github.com/vmware/repository-service-tuf/issues/188

### References

_No response_

### Code of Conduct

kairoaraujo:
This improves the performance of updating Targets in the SQL DB.

Previously, we used to update the targets one by one in the DB and this commit made it in bulk.

In publishing new targets, we used to persist in the hashed bins with new targets to the backend storage, update the snapshot, and update the timestamp followed by the targets one by one in the SQL DB.

The last step was improved doing the update of all published targets at once.

Closes #176

### DIFF
--- a/repository_service_tuf_worker/models/targets/crud.py
+++ b/repository_service_tuf_worker/models/targets/crud.py
@@ -119,19 +119,20 @@ def update(
     return target
 
 
-def update_to_published(
-    db: Session, target: models.RSTUFTargets
-) -> models.RSTUFTargets:
+def update_to_published(db: Session, paths: List[str]) -> None:
     """
     Update Target to `published` to `True`.
     """
-    target.published = True
-    target.last_update = datetime.now()
-    db.add(target)
-    db.commit()
-    db.refresh(target)
 
-    return target
+    db.query(models.RSTUFTargets).filter(
+        models.RSTUFTargets.path.in_(paths)
+    ).update(
+        {
+            models.RSTUFTargets.published: True,
+            models.RSTUFTargets.last_update: datetime.now(),
+        }
+    )
+    db.commit()
 
 
 def update_action_remove(
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -263,7 +263,7 @@ class MetadataRepository:
     def _update_timestamp(
         self,
         snapshot_version: int,
-        db_targets: Optional[List[targets_models.RSTUFTargets]] = None,
+        db_targets: Optional[List[str]] = None,
     ) -> Metadata[Timestamp]:
         """
         Loads 'timestamp', updates meta info about passed 'snapshot'
@@ -284,8 +284,7 @@ class MetadataRepository:
 
         # TODO review if here is the best place to change the status in DB
         if db_targets:
-            for db_target in db_targets:
-                targets_crud.update_to_published(self._db, db_target)
+            targets_crud.update_to_published(self._db, db_targets)
 
         return timestamp
 
@@ -442,7 +441,7 @@ class MetadataRepository:
             # initialize the new snapshot targets meta and published targets
             # from DB SQL
             new_snapshot_meta: List[Tuple(str, int)] = []
-            db_published_targets: List[targets_models.RSTUFTargets] = []
+            db_published_targets: List[str] = []
             for _, rolename in unpublished_roles:
                 # get the unpublished targets for the delegated, it will be use
                 # to update the database when Snapshot and Timestamp is
@@ -451,7 +450,7 @@ class MetadataRepository:
                     db=self._db, rolename=rolename
                 )
                 logging.debug(f"{rolename}: New targets #: {len(db_targets)}")
-                db_published_targets += db_targets
+                db_published_targets += [target.path for target in db_targets]
 
                 # load the delegated targets role, clean the targets and add
                 # a new meta from the SQL DB.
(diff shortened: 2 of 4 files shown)


repository_service_tuf_worker/models/targets/crud.py
```diff
@@ -119,19 +119,20 @@ def update(
...
-    return target
+    db.query(models.RSTUFTargets).filter(
+        models.RSTUFTargets.path.in_(paths)
+    ).update(
+        {
+            models.RSTUFTargets.published: True,
+            models.RSTUFTargets.last_update: datetime.now(),
+        }
+    )
+    db.commit()
```
MVrachev, line 135: I see you also did `db.refresh(target)` before.
This is not required anymore because `db.update()` does that job right?

repository_service_tuf_worker/repository.py
```diff
@@ -284,8 +284,7 @@ def _update_timestamp(
 
         # TODO review if here is the best place to change the status in DB
```
MVrachev, line 285: You mentioned this `TODO` in the issue you have linked.
Can we remove it or this is question is still open?

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
```diff
@@ -211,13 +211,24 @@ def test_update(self, monkeypatch):
         assert fake_datetime.now.calls == [pretend.call()]
 
     def test_update_to_published(self, monkeypatch):
-
+        test_targets = ["path/file1", "path.file2"]
+        crud.models.RSTUFTargets = pretend.stub(
+            path=pretend.stub(
+                in_=pretend.call_recorder(lambda *a: test_targets)
+            ),
+            published="published",
```
MVrachev, line 219: Nit: I think the `published` value should be `True` as in the `update_to_published` you are updating it to `True`.

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
```diff
@@ -211,13 +211,24 @@ def test_update(self, monkeypatch):
...
+            last_update="last_update",
```
MVrachev, line 220: Shouldn't you use the `fake_time` from below as you are updating it to `datetime.now()` which you mock below?

repository_service_tuf_worker/models/targets/crud.py
kairoaraujo: The `db.refresh()` is required when you will return the object, not to return it with the initial value.
We no long return the object because we don't give it to the function, only the paths.

repository_service_tuf_worker/repository.py
kairoaraujo: I would keep the question still open. We must use it more to see if it is the right/best place.

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
kairoaraujo: If I do it, I will need to change the assert by something less readable.

```python
        assert mocked_update.update.calls == [
            pretend.call({"published": True, "last_update": fake_time})
        ]
```
to
```python
        assert mocked_update.update.calls == [
            pretend.call({"True": True, fake_time: fake_time})
        ]
```

This is because we are not controlling/testing what is inside the SQLAlchemy.
Which one do you prefer?

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
kairoaraujo: See my comment above.

tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
MVrachev: I prefer the first. I got what you do here.
By stubbing the value of `crud.models.RSTUFTargets.published` you actually replace what value will be placed when you access it and given that the `update` API call accepts a `Dict[str, Any]` it should be a string.

MVrachev (approved):

MVrachev:
I am merging as I agree with the changes.
If you have any other improvements to make based on our discussion you can do them in another pr.

PR #187, opened 2023-01-26
Skip refresh settings on subtasks

MVrachev:
There is no point in refreshing the repository settings when we launch
a subtask from a main task that has already updated the repository
settings.
That's what happens when we launch a new "publish_targets" from the
"repository._send_publish_targets_task()" function.

By default this optional argument "refresh_settings" is false to
make sure we don't break any of the current components launching a
worker task.

Close https://github.com/vmware/repository-service-tuf-worker/issues/178

### DIFF
--- a/app.py
+++ b/app.py
@@ -75,14 +75,14 @@ def repository_service_tuf_worker(
     refresh_settings: Optional[bool] = True,
 ):
     """
-    Accepts Celery tasks with an action to be executed.
+    Repository Service for TUF Metadata Worker main Celery consumer.
 
     Args:
         action: which action to be executed by the task.
         payload: data that will be given to the action.
         refresh_settings: whether to refresh repository settings.
     """
-    if refresh_settings:
+    if refresh_settings is True:
         repository.refresh_settings(worker_settings)
 
     repository_action = getattr(repository, action)
--- a/app.py
+++ b/app.py
@@ -75,7 +75,12 @@ def repository_service_tuf_worker(
     refresh_settings: Optional[bool] = True,
 ):
     """
-    Repository Service for TUF Metadata Worker
+    Accepts Celery tasks with an action to be executed.
+
+    Args:
+        action: which action to be executed by the task.
+        payload: data that will be given to the action.
+        refresh_settings: whether to refresh repository settings.
     """
     if refresh_settings:
         repository.refresh_settings(worker_settings)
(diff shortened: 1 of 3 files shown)


tests/unit/test_app.py
```diff
@@ -47,12 +47,29 @@ def test_repository_service_tuf_worker_no_payload(self):
             test_action=pretend.call_recorder(lambda *a, **kw: True),
         )
 
+        result = app.repository_service_tuf_worker("test_action")
+
+        assert result is True
+        assert app.repository.test_action.calls == [
+            pretend.call(),
+        ]
```
MVrachev, line 55: Git showed those as changed as I haven't touched those lines.
Instead I have added those two lines in the next unit test:
`test_repository_service_tuf_worker_refresh_settings_false`.

app.py
```diff
@@ -69,12 +69,22 @@ class status(Enum):
...
 def repository_service_tuf_worker(
-    self, action: str, payload: Optional[Dict[str, Any]] = None
+    self,
+    action: str,
+    payload: Optional[Dict[str, Any]] = None,
+    refresh_settings: Optional[bool] = True,
 ):
     """
-    Repository Service for TUF Metadata Worker
+    Accepts Celery tasks with an action to be executed.
```
kairoaraujo, line 78:
```suggestion
    Repository Service for TUF Metadata Worker main Celery consumer.
    
    Consume tasks from the configure broker.
```

tests/unit/test_app.py
kairoaraujo: It's a bug. My PR also shows the same.

kairoaraujo (changes requested):

app.py
```diff
@@ -69,12 +69,22 @@ class status(Enum):
...
+
+    Args:
+        action: which action to be executed by the task.
+        payload: data that will be given to the action.
+        refresh_settings: whether to refresh repository settings.
     """
-    repository.refresh_settings(worker_settings)
+    if refresh_settings:
```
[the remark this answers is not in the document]
MVrachev: Totally forgot about this, good catch.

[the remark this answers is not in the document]
MVrachev: I think because I still I had this thinking from a typed language that when I say that an argument is `boolean` then it truly is. :D

kairoaraujo (approved):

PR #188, opened 2023-01-27
fix: remove wrong entry in the file

kairoaraujo:
RSTUF is MIT licensed.
Removed some wrong entry in the file.


joshuagl (approved):
Thanks

PR #194, opened 2023-01-30
fix `make requirements` and update requirements

### LINKED ISSUE #185: Other: make requirements does not update Pipfile.lock when Pipfile has been manually
### What happened?

Running make requirements does not update the requirement files (requirements.txt and requirements-dev.txt) correctly if we have manually edited the Pipfile. This could happen if we are manually pinning a dependency or specifying some optional dependency.
Locking first could protect us from having outdated requirement files if we manually edit the Pipfile and would be a nice-to-have extra precaution.

### What steps did you take?

1. Update the Pipfile by adding a new package
2. Run make requirements to update the requirement files

We expect the requirements files to include the latest changes made to the Pipfile.
This does not happen since we need to lock before running:

```
pipenv requirements > requirements.txt
pipenv requirements --dev > requirements-dev.txt
```
so we could instead have:

```
pipenv lock
pipenv requirements > requirements.txt
pipenv requirements --dev > requirements-dev.txt
```
This is because the Pipfile.lock is used to generate the requirement files

### Relevant log output

_No response_

### Code of Conduct

### References

_No response_

### Code of Conduct

kairoaraujo:
- fix `make requirements`
- update worker requirements
  - remove pinned development depencies from Pipfile
- update `.pre-commit-config.yaml` following `requirements-dev.txt`

Closes #185


KAUTH (commented):
- In https://github.com/vmware/repository-service-tuf-worker/blob/main/.github/workflows/ci.yml#L32 we could run `make tests`

Pipfile
```diff
@@ -9,7 +9,7 @@ watchdog = "*"
 cryptography = "*"
 pynacl = "*"
 redis = "*"
-tuf = "==2.0.0"
+tuf = "*"
 dynaconf = {extras = ["ini"], version = "*"}
 supervisor = "*"
 securesystemslib = "*"
```
KAUTH, line 15: I believe at least in `repository_service_tuf_worker/services/keyvault/local.py` (`decrypt_key`, `encrypt_key`) we use the cryptography feature so we might want to install the `crypto` optional dependency:
```suggestion
securesystemslib = {extras = ["crypto"], version = "*"}
```

Pipfile
```diff
@@ -19,71 +19,71 @@ alembic = "*"
...
+coverage = "*"
+black = "*"
+isort = "*"
+mypy = "*"
+pytest = "*"
+pretend = "*"
+amqp = "*"
+async-timeout = "*"
+attrs = "*"
+billiard = "*"
```
KAUTH, line 33: Is the `billiard` package used? (it's a dependency of `celery`, but I don't think it's needed as a dev dep)

Pipfile
```diff
@@ -19,71 +19,71 @@ alembic = "*"
...
+celery = "*"
```
KAUTH, line 34: `celery` is probably not needed as a dev dep. I think some dev packages were added as cross-dependencies from the main dependenices

Pipfile
```diff
@@ -19,71 +19,71 @@ alembic = "*"
...
+click = "*"
+configobj = "*"
+cryptography = "*"
+deprecated = "*"
+distlib = "*"
+dynaconf = {version = "*", extras = ["ini"]}
+filelock = "*"
+idna = "*"
+iniconfig = "*"
+kombu = "*"
```
KAUTH, line 50: To list all the devs dependencies that don't seem to be used and should be removed:
deps of `celery`:
- `kombu`
- `click-didyoumean`
- `click-plugins`
- `click-repl`
- `prompt-toolkit`

To list them all here, deps of `pytest`:
- `iniconfig`
- `pluggy`

Same for:
- `certifi`
- `charset-normalizer`
- `click`
- `configobj `?
- `configobj `?
- `deprecated` (dep of `redis`)
- `distlib`
- `dynaconf = {version = "*", extras = ["ini"]}`
- `filelock`
- `idna`
- `mccabe`
- `mypy-extensions`
- `pathspec`
- `platformdirs`
- `pyparsing`
- `pytz`
- `py`
- `pycodestyle`
- `pycparser`
- `pyflakes`
- `pynacl`
- redis = "*"
requests = "*"
securesystemslib = "*"
six = "*"
toml = "*"
tomli = "*"
tuf = "*"
typing-extensions = "*"
urllib3 = "*"
vine = "*"
virtualenv = "*"
watchdog = "*"
wcwidth = "*"
wrapt = "*"

KAUTH (changes requested):

Pipfile
MVrachev: What will help us to have `crypto` as an optional dependency?

Pipfile
```diff
@@ -19,71 +19,71 @@ alembic = "*"
...
-sphinxcontrib-serializinghtml = "==1.1.5"
+tox = "*"
+flake8 = "*"
+coverage = "*"
+black = "*"
+isort = "*"
+mypy = "*"
+pytest = "*"
+pretend = "*"
+amqp = "*"
```
MVrachev, line 30: Why did we pin the requirements before and why we would not pin them now?
Does this mean that there won't be so many regular prs updating the build dependencies?

kairoaraujo: It was pinned (`requirements-dev.txt`) by mistake at some point.
We don't pin them in the `Pipenv` only when some incompatibility requires it. The `Pipenv.lock` is the pin control.
It is pinned automatically in the `requirements*.txt` file (`make requirements`) locks and pin it to offer reproducibility in each release.
The `requirement.txt` are used to release the container images in the `Dockerfile`

Pipfile
kairoaraujo: We already use `pynacl` https://github.com/secure-systems-lab/securesystemslib#installation
Not sure if we need crypto. 🤔

Pipfile
kairoaraujo: Thank you for spotting it @KAUTH 
I cleaned the Pipfile.

Pipfile
KAUTH: My comment was based on https://github.com/vmware/repository-service-tuf-cli/issues/131 which we also required crypto when installing `securesystemslib`

Pipfile
kairoaraujo: @KAUTH ready for another review :)

KAUTH (changes requested):

Pipfile
```diff
@@ -4,12 +4,11 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-celery = "*"
```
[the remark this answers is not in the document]
kairoaraujo: Ouch something were wrong. Thank you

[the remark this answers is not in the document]
KAUTH: @kairoaraujo btw how come none of our tests caught that Celery was missing? I think they were passing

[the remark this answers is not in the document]
kairoaraujo: Not sure. Maybe some circular dependency. I tried to investigate but I couldn't find it.

MVrachev (approved):
I agree with the changes, but as @KAUTH is already involved much in the review process I won't merge until he gives his approval as well.

KAUTH:
> I agree with the changes, but as @KAUTH is already involved much in the review process I won't merge until he gives his approval as well.

Just a small concern about [this](https://github.com/vmware/repository-service-tuf-worker/pull/194#discussion_r1092262576) comment.

KAUTH:
> > cryptography
> 
> The traceback you did @KAUTH is very nice! But we explicitly added `cryptography` (as we did for `pynacl`).
> 
> https://github.com/vmware/repository-service-tuf-worker/blob/bfdbfc59083fa5156baad37e73ca8931d371f754/Pipfile#L8
> 
> https://github.com/vmware/repository-service-tuf-worker/blob/bfdbfc59083fa5156baad37e73ca8931d371f754/Pipfile#L9

Aha, indeed. The only drawback is that we specify, for example, `cryptography = "*"`, while if we added the requirement via the extra dependency of `securesystemslib` we would have the version specifications that `securesystemslib` needs, in this example currently that would be: `cryptography >= 37.0.0` (https://github.com/secure-systems-lab/securesystemslib/blob/master/pyproject.toml#L46, https://github.com/secure-systems-lab/securesystemslib/blob/master/requirements.txt#L8). We miss that extra bit of information that `securesystemslib` imposes. This could be something minor, but it would be nice to have since these libraries are only necessary because of `securesystemslib`.

kairoaraujo:
> Aha, indeed. The only drawback is that we specify, for example, `cryptography = "*"`, while if we added the requirement via the extra dependency of `securesystemslib` we would have the version specifications that `securesystemslib` needs, in this example currently that would be: `cryptography >= 37.0.0` (https://github.com/secure-systems-lab/securesystemslib/blob/master/pyproject.toml#L46, https://github.com/secure-systems-lab/securesystemslib/blob/master/requirements.txt#L8). We miss that extra bit of information that `securesystemslib` imposes. This could be something minor, but it would be nice to have since these libraries are only necessary because of `securesystemslib`.

You have a good point. 
It is an amazing way of thinking about dependencies. Thank you @KAUTH 👏

KAUTH (approved):
LGTM!

MVrachev (approved):

PR #197, opened 2023-02-01
Monitor publish tasks sub task and remove automatic job publishing targets

### LINKED ISSUE #190: Enhancement: Improve the task monitoring the sub-task, to have a better update for task

Improve the task monitoring the sub-task, to have a better update for the user.

The user adds/removes target(s) sending a request and receives a `task id` to monitor the status and progress.
This task in the worker triggers a sub-task, the task should monitor this sub-task to update the task.

Note: Worker N, when N can be any Worker of multiples

```mermaid
sequenceDiagram
    participant Broker/Result Backend
    participant Worker N
    participant DB
    participant Storage Service
    Worker N-->>Broker/Result Backend: Get the Task to add/remove
    Worker N->>DB: Insert/update targets
    Worker N->>Broker/Result Backend: Update add/remove Task 
    Worker N->>Broker/Result Backend: submit sub-task `publish_targets `
    Worker N-->>Broker/Result Backend: Get the task `publish_targets `
    critical sub-task `publish_targets`
      Worker N->> Storage Service: Update `bins-{hash}`, Update Snapshot, Update Timestamp
      Worker N->> DB: Update SQL DB
    end
    Worker N->>Broker/Result Backend: Update add/remove Task FINISH
```

### References

- https://github.com/vmware/repository-service-tuf-worker/issues/175

### Code of Conduct

kairoaraujo:
When adding/removing targets, it triggers a new job to publish the
targets.

This subtask could fail for some reason, and we don't monitor it.

This commit adds the monitoring for this task and will give details
to the user once it fails the parent task (add/remove targets).

Closes #190

The monitoring is done in the `_update_task` as part of checks.

As discussed in issue https://github.com/vmware/repository-service-tuf-worker/issues/175, we removed the automatic job that
publishes targets every minute.

The publish targets are done to persist in the RSTUF Worker backend
storage of the updated metadata with the new targets.

1. We will give full control over the failure to add/remove targets
to the user. They can decide to reprocess the entire flow (add/remove
targets), or they can trigger publish targets.
2. The automatic job publishes targets creates confusion when and who
publishes the targets if the automatic job wins in the race condition
with add/remove targets.
3. For the feature load/import data it can slow down the process and
it would require an way to disable the automatic jobs during this
process, adding a lot of confusion.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -25,6 +25,8 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import redis
 from celery.app.task import Task
+from celery.exceptions import ChordError
+from celery.result import AsyncResult, states
 from securesystemslib.exceptions import StorageError  # type: ignore
 from securesystemslib.signer import SSlibSigner  # type: ignore
 from tuf.api.metadata import (  # noqa
@@ -321,6 +323,7 @@ class MetadataRepository:
     def _update_task(
         self,
         bin_targets: Dict[str, List[targets_models.RSTUFTargets]],
+        subtask: AsyncResult,
         update_state: Task.update_state,
     ):
         """
@@ -331,16 +334,21 @@ class MetadataRepository:
         logging.debug(f"Waiting roles to be published {list(bin_targets)}")
 
         def _update_state(
+            state: states,
             bin_targets: Dict[str, List[targets_models.RSTUFTargets]],
             completed_roles: List[str],
+            exc_type: Optional[str] = None,
+            exc_message: Optional[List[str]] = None,
         ):
             update_state(
-                state="RUNNING",
+                state=state,
                 meta={
                     "published_roles": completed_roles,
                     "roles_to_publish": f"{list(bin_targets.keys())}",
                     "status": "Publishing",
                     "last_update": datetime.now(),
+                    "exc_type": exc_type,
+                    "exc_message": exc_message,
                 },
             )
 
@@ -355,9 +363,23 @@ class MetadataRepository:
                 if len(targets) == 0:
                     logging.debug(f"Update: {role_name} completed")
                     completed_roles.append(role_name)
+            if subtask.status == states.FAILURE:
+                exc_type = subtask.result.__class__.__name__
+                exc_message = list(subtask.result.args)
+                _update_state(
+                    states.FAILURE,
+                    bin_targets,
+                    completed_roles,
+                    exc_type=exc_type,
+                    exc_message=exc_message,
+                )
+                raise ChordError(
+                    f"Failed to execute {subtask.task_id}: "
+                    f"{exc_type} {exc_message}"
+                )
 
             if sorted(completed_roles) != sorted(list(bin_targets)):
-                _update_state(bin_targets, completed_roles)
+                _update_state("RUNNING", bin_targets, completed_roles)
                 time.sleep(3)
             else:
                 break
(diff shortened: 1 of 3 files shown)


kairoaraujo:
Requires #194 for the Lint issue in the checks.

repository_service_tuf_worker/repository.py
```diff
@@ -355,9 +363,23 @@ def _update_state(
...
+                    exc_message=exc_message,
+                )
+                raise ChordError(
+                    f"Failed to execute {subtask.task_id}: "
+                    f"{exc_type} {exc_message}"
+                )
 
             if sorted(completed_roles) != sorted(list(bin_targets)):
-                _update_state(bin_targets, completed_roles)
+                _update_state("RUNNING", bin_targets, completed_roles)
```
MVrachev, line 382: I specifically opened the `Celery.result.states` module and I don't find a state called `RUNNING`.
Probably here it could be a custom string, but I just wanted to mention that.

repository_service_tuf_worker/repository.py
```diff
@@ -531,8 +553,8 @@ def add_targets(
 
             bin_targets[bins_name].append(db_target)
 
-        self._send_publish_targets_task(task_id)
-        self._update_task(bin_targets, update_state)
+        subtask = self._send_publish_targets_task(task_id)
```
MVrachev, line 556: We are sure we would have the final state of the main task or it could be some intermediate state?

repository_service_tuf_worker/repository.py
kairoaraujo: Yes, there is no running :(

repository_service_tuf_worker/repository.py
kairoaraujo: If the `.apply_async()` fails, it raises an exception. So the main task will be`FAILURE`. The most common case is for some reason the access to the Broker fails.
So, if not, it will return the object.
The `subtask` Celery task object.
This object contains attribute implementations containing built-in "callbacks" to retrieve the task from celery.
For example, retrieve task status, results, etc.
You can use the return from `.apply_async()` or you can build the object using `celery.Task.AsyncResult(<taskid>)`

repository_service_tuf_worker/repository.py
MVrachev: This means that here we could have a custom string as a state, right?
This is correct?

MVrachev (approved):
I think we need to merge https://github.com/vmware/repository-service-tuf-worker/pull/194 and then rebuild it again.

MVrachev:
You can rebase this pr and get the fix from https://github.com/vmware/repository-service-tuf-worker/pull/194 as it's merged.

repository_service_tuf_worker/repository.py
kairoaraujo: Yes, we have a custom string as state RUNNING.
I notice that we declare them here, andt it can make it more consistent.

https://github.com/vmware/repository-service-tuf-worker/blob/fda0e273e974acb6c957d35b4dfe9a43827c9654/app.py#L30-L36

I will not add it to this PR. Otherwise, it will change the focus and the primary goal.

The new issue is:

Make the state more consistent.

1. Rename from `status` to `STATE` as a constant
https://github.com/vmware/repository-service-tuf-worker/blob/fda0e273e974acb6c957d35b4dfe9a43827c9654/app.py#L30
2. Move it to `repository_service_tuf_worker.__init__`
3. Use the Celery `states` instead new strings, keep the strings only for RSTUF custom
```python
 class STATE(Enum): 
     RECEIVED = states.RECEIVED 
     PRE_RUN = "PRE_RUN" 
     RUNNING = "RUNNING" 
     SUCCESS = states.SUCCESS
     UNKNOWN = "UNKNOWN" 
     FAILURE = states.FAILURE
``` 
4. Apply the refactoring to all usage (including UT)

kairoaraujo: Issue created

kairoaraujo:
@MVrachev, Rebased.

MVrachev (approved):

PR #203, opened 2023-02-07
Accept "publish_targets" in add/remove targets

### LINKED ISSUE #198: feature: skip the `publish_targets` when field is `False` in the payload

If in the `publish_targets` there is a field with value `False`, skip the `_send_publish_targets_task` sub-task in the add/remove targets.
This field was discussed in #

### References

- #175

### Code of Conduct

MVrachev:
There is a new field that will be introduced in the RSTUF add and remove
API calls named "publish_targets".
This flag can be used by users to tell whether to start a subtask
publishing the latest targets changes.
By default, this flag will be enabled.

Close #198 

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -323,8 +323,8 @@ class MetadataRepository:
     def _update_task(
         self,
         bin_targets: Dict[str, List[targets_models.RSTUFTargets]],
-        subtask: AsyncResult,
         update_state: Task.update_state,
+        subtask: Optional[AsyncResult] = None,
     ):
         """
         Updates the 'RUNNING' state with details if the meta still not
@@ -363,7 +363,7 @@ class MetadataRepository:
                 if len(targets) == 0:
                     logging.debug(f"Update: {role_name} completed")
                     completed_roles.append(role_name)
-            if subtask.status == states.FAILURE:
+            if subtask is not None and subtask.status == states.FAILURE:
                 exc_type = subtask.result.__class__.__name__
                 exc_message = list(subtask.result.args)
                 _update_state(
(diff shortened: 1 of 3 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -553,8 +553,11 @@ def add_targets(
 
             bin_targets[bins_name].append(db_target)
 
-        subtask = self._send_publish_targets_task(task_id)
-        self._update_task(bin_targets, subtask, update_state)
+        # If publish_targets doesn't exists it will be True by default.
+        publish_targets = payload.get("publish_targets", True)
+        if publish_targets is True:
+            subtask = self._send_publish_targets_task(task_id)
+            self._update_task(bin_targets, subtask, update_state)
```
kairoaraujo, line 560: The `self._update_task` is the job that keeps updating the main task progress (`update_state`, not only the `sub_task` progress.
This will require some changes:
1. The `self._update_task` needs to be moved outside of the if the condition
2. In the `self._update_task` the parameter `subtask` needs to be `Optional[AsyncResult] = None`

kairoaraujo (changes requested):
Thank you @MVrachev !

repository_service_tuf_worker/repository.py
```diff
@@ -323,8 +323,8 @@ def _get_path_succinct_role(self, target_path: str) -> str:
     def _update_task(
         self,
         bin_targets: Dict[str, List[targets_models.RSTUFTargets]],
-        subtask: AsyncResult,
         update_state: Task.update_state,
+        subtask: Optional[AsyncResult] = None,
```
MVrachev, line 327: I had to replace where the `subtask` is because I give it a default value and arguments with default values cannot be before arguments without a default value.

kairoaraujo (approved):

PR #206, opened 2023-02-14
Add dockerignore file

kapsalis:
Add dockerignore file to exclude big size directories when sending build context to the Docker deamon. These dirs are not needed during the docker image build process, and they slow down the start of the build process.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.tox/
+venv/
+.idea/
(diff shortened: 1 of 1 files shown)


MVrachev (approved):

MVrachev:
We had some discussions for this file in https://github.com/vmware/repository-service-tuf-api/pull/258.

Issue #202, opened 2023-02-07
Task: validate keys type and algorithm 


MVrachev:
### What is the task about?

Validate the key types and algorithms during the bootstrap process.
In order to do this we can:
1.  import the securesystemslib [KEY_FOR_TYPE_AND_SCHEME](https://github.com/secure-systems-lab/securesystemslib/blob/34e50cb059ea630e7a808dca7c1744b20f2d8ce4/securesystemslib/signer/__init__.py#L30) dict
2. Update the `KEY_FOR_TYPE_AND_SCHEME` by cleaning its content with `dict.clean()` and instead adding our supported algorithms
3. During the bootstrap ceremony when loading the public keys stored in `root.singed.keys` we should do it with the [Key.from_dict](https://github.com/secure-systems-lab/securesystemslib/blob/34e50cb059ea630e7a808dca7c1744b20f2d8ce4/securesystemslib/signer/_key.py#L74) API call to use the validation inside securesystemslib. 

PS: This validation will be required when the rotating keys feature is implemented as well.

### Parent feature

https://github.com/vmware/repository-service-tuf/issues/208

### References

https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/signer/_key.py
https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/signer/__init__.py
https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/signer/_signer.py

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MVrachev:
I can work on this one.

Gitjay11:
I'm working on this issue, will raise a PR soon.

PR #210, opened 2023-02-22
Remove IKeyVault write functionality and refactor IKeyVault.get()

MVrachev:
We have decided that our keyvaults will be read-only and we no longer
need the "put" method.
Instead to load the online key (which is used for all roles except root)
we use environmental variables.

We have also decided to change the purpose of "IKeyVault.get()" and
instead of returning a "Key" object make it to return a signer and
because of that, I renamed it to "get_signer()". 

We can do that because we want to load the private
key inside the "get()" call anyway which is used for signing.

Finally, I have refactored "get_singer()" to generate a signer for the
online key that will be shared for all roles except root.

Closes https://github.com/vmware/repository-service-tuf-worker/issues/170


repository_service_tuf_worker/repository.py
```diff
@@ -218,8 +217,7 @@ def _sign(self, role: Metadata, role_name: str) -> None:
         for top-level roles.
         """
         role.signatures.clear()
-        for key in self._key_storage_backend.get(role_name):
-            signer = SSlibSigner(key)
+        for signer in self._key_storage_backend.get_signer():
```
lukpueh, line 220: I don't think you want to iterate over the returned `Signer`.

repository_service_tuf_worker/services/keyvault/local.py
```diff
@@ -76,33 +82,21 @@ def settings(cls) -> List[ServiceSettings]:
...
+            key_info: Dict[str, Any] = import_privatekey_from_file(
+                self._online_key_name,
+                self._online_key_type,
+                self._online_key_password,
             )
-
-        self.keyvault.store[rolename.upper()] = key_vault_data
-        data = self.keyvault.as_dict(env=self.keyvault.current_env)
-        loaders.write(self._secrets_file, DynaBox(data).to_dict())
+            return SSlibSigner(key_info)
```
lukpueh, line 93: I still think your RSTUF KeyVault abstraction over the secureystemslib signer interface should take a public key and use from_priv_key_uri. Then you'd be able to reuse the same code for any Signer implementation.

I understand that you wanted to avoid passing in a public key,  which is possible for SSlibSigner, but will be harder for something like a cloud signer (e.g. GCPSigner), where the signer doesn't have direct access to the key material.

Plus it should not be a problem to get the public key for the signer. You can even grab it from root or top-level targets. :)


cc @jku

repository_service_tuf_worker/interfaces.py
```diff
@@ -48,15 +49,8 @@ def settings(cls) -> List[ServiceSettings]:
...
-    def get(self, rolename: List[str]) -> Dict[str, Any]:
-        """Return a key from specific rolename."""
-        pass  # pragma: no cover
-
-    @abstractmethod
-    def put(self, file_object: str, filename: str) -> None:
-        """
-        Stores file object with a specific filename.
-        """
+    def get_signer(self) -> Signer:
```
kairoaraujo, line 52:
```suggestion
    def get(self) -> Signer:
```

We don't need to rename the existing method.
If we define that returns, a Singer makes ok. 
IMHO, keep the name `get` aligned with the other Interface (`IStorage`) with the method `get`.

As it is an API (GET, PUT, DELETE, etc.) will look better.

repository_service_tuf_worker/services/keyvault/local.py
kairoaraujo: I deleted by mistake my comment

Each RSTUF KeyVault abstraction (service) will define its [required parameters.](https://repository-service-tuf.readthedocs.io/en/latest/guide/repository-service-tuf-worker/Docker_README.html#required-rstuf-keyvault-backend) (required or optional)

It could be difficult to predict all the SaaS/Vault required parameters.

I will simulate different examples here:

* RSTUF_KEYVAULT_BACKEND: LocalKeyVault
  - RSTUF_LOCAL_KEYVAULT_PATH: /volume/path/
  - RSTUF_LOCAL_KEYVAULT_NAME: rstuf-online.key
  - RSTUF_LOCAL_KEYVAULT_TYPE: rsa
  - RSTUF_LOCAL_KEYVAULT_PASSWORD: /run/secrets/RSTUF_ONLINE_KEY

* RSTUF_KEYVAULT_BACKEND: GCPKM
  - RSTUF_GCPKM_KEYVAULT_PUBLIC_KEY: "-----BEGIN PUBLIC KEY----- ..."
  - RSTUF_GCPKM_KEYVAULT_PUBLIC_URI: "gcp-kms://projects/*/locations/*/keyRings/*/cryptoKeys/*"
 
* RSTUF_KEYVAULT_BACKEND: AWSKMS
  - RSTUF_AWSKMS_KEYVAULT_KEY_ID: "arn:aws:kms:<region>:<account-id>:key/<key-id>"

jku: I'm definitely not telling you what makes sense, I'm not familiar with this code, but since lukas pinged... This is how I would handle your examples:
* GCP: Use GCPSigner
* AWS: Implement AWSSigner in securesystemslib (with a private key uri that contains the AWS keyid), then use it
* LocalKeyVault: Implement own signer (maybe in application code?), with a private_key_uri containing all of the constant arguments you mentioned. If the arguments are not constant (or are secrets) you read them from a location that is constant and can be embedded in private_key_uri -- this is what the other signers do as well

lukpueh: My main point here is about the public key. All current signer implementations need it, and RSTUF has it (via root metadata). So why not use that?

I don't think it's necessary to find a custom solution for every signer, such as passing it via a config setting (`RSTUF_GCPKM_KEYVAULT_PUBLIC_KEY`), or working around it by inferring it from the private key material (`SSlibSigner`).

repository_service_tuf_worker/interfaces.py
MVrachev: Okay, I don't have a strong opinion on this one, so I will return it to `get()`.

repository_service_tuf_worker/services/keyvault/local.py
MVrachev: I see the benefit of passing the public key and using `from_priv_key_uri` directly.
For us, the biggest benefit is that we can easily have #202 implemented by redefining `KEY_FOR_TYPE_AND_SCHEME` however we want and using `Key. from_dict()`

MVrachev:
I made multiple changes and some of them where optional as the one changing the `LocalStorage.get()` interface, but because I had to use it I just noticed it was wrong and fixed it right away.
I sent the public key information as suggested.
Ready for another review @kairoaraujo and @lukpueh.

tox.ini
```diff
@@ -14,9 +14,9 @@ setenv =
     RSTUF_LOCAL_STORAGE_BACKEND_PATH = ./data-test/s
     RSTUF_KEYVAULT_BACKEND = LocalKeyVault
     RSTUF_LOCAL_KEYVAULT_PATH = ./data-test/kv
-    RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_NAME = ./data-test/kv/online_key
-    RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_PASSWORD = password
-    RSTUF_LOCAL_KEYVAULT_ONLINE_KEY_TYPE = ed25519
+    RSTUF_LOCAL_KEYVAULT_KEY_NAME = ./data-test/kv/online_key
+    RSTUF_LOCAL_KEYVAULT_KEY_PASSWORD = password
+    RSTUF_LOCAL_KEYVAULT_KEY_TYPE = ed25519
```
kairoaraujo, line 19: We should remove this one (to test as optional).
It will fail all tests because the `app.py` initialization will fail. Besides we defined the parameters as optional the `LocalKeyVault`, it misses the condition in `refresh_settings` (you can blame me for this) to not fetch the configuration from environment variable/Dynaconf.

The required changes are:

1) Storage backend

From:
https://github.com/vmware/repository-service-tuf-worker/blob/cc13a730760f8dc131539771d6e2854432ca0d78/repository_service_tuf_worker/repository.py#L149-L152

To:

```python
 storage_kwargs = { 
     s.argument: settings.store[s.name] 
     for s in settings.STORAGE_BACKEND.settings() 
     if settings.store.get(s.name)
 } 
```

2) Keyvault Backend


From:

https://github.com/vmware/repository-service-tuf-worker/blob/cc13a730760f8dc131539771d6e2854432ca0d78/repository_service_tuf_worker/repository.py#L192-L195

To:
 ```python
keyvault_kwargs = { 
     s.argument: settings.store[s.name] 
     for s in settings.KEYVAULT_BACKEND.settings()
     if settings.store.get(s.name)
 } 
```

**Review by kairoaraujo — CHANGES_REQUESTED** [review]
Hi @MVrachev, that is excellent work.

I made some comments after testing it on my development/test environment.

I found a problem: the container will not fail at the start if the key doesn't exist.
It could lead the RSTUF users in deployment troubleshooting by a wrong key name, password, corrupted file, permission error, etc. 

The user will receive this error only when they bootstrap and at the first signing by the online key. It would be better if it could fail at the start.

The big deal is that we require the public key, and we don't as it to the user, so we get it from the Root Metadata, but before a bootstrap, we don't have the metadata either.

If we have a later failure, only the user does the bootstrap, which might give a bad user experience IMHO. Even overcomplicating the error, if the user just added a wrong `key_name`, it will be discovered after the bootstrap when we have the public key from the Root Metadata.

**Comment by MVrachev** [comment]
We agreed with @kairoaraujo that I will remove `LocalKeyVault.path`, rename `LocalKeyVault.key_name` to `key_path` and make it as required.

**Comment by MVrachev** [comment]
I had to rebase because I needed the changes from #222.
The changes in the latest iteration after the reviews are from "@MVrachev
[Remove unnecessary LocalKeyVault arguments](https://github.com/vmware/repository-service-tuf-worker/pull/210/commits/d408966aa5a11b4301dd2bfa94b68cfc80e7df0c)" commit.

**Review by kairoaraujo — CHANGES_REQUESTED** [review]

File: docker-compose.yml [hunk 5]
```diff
@@ -76,15 +73,15 @@ services:
       - RSTUF_STORAGE_BACKEND=LocalStorage
       - RSTUF_KEYVAULT_BACKEND=LocalKeyVault
       - RSTUF_LOCAL_STORAGE_BACKEND_PATH=/var/opt/repository-service-tuf/storage
-      - RSTUF_LOCAL_KEYVAULT_PATH=/var/opt/repository-service-tuf/keystorage
+      - RSTUF_LOCAL_KEYVAULT_KEY_PATH=/opt/repository-service-tuf-worker/tests/files/key_storage/online.key
```
kairoaraujo, line 76:
```suggestion
      - RSTUF_LOCAL_KEYVAULT_PATH=/opt/repository-service-tuf-worker/tests/files/key_storage/online.key
```

docs/source/guide/Docker_README.md
```diff
@@ -29,7 +29,6 @@ docker run --env="RSTUF_WORKER_ID=worker1" \
     --env="RSTUF_STORAGE_BACKEND=LocalStorage" \
     --env="RSTUF_LOCAL_STORAGE_BACKEND_PATH=storage" \
     --env="RSTUF_KEYVAULT_BACKEND=LocalKeyVault" \
-    --env="RSTUF_LOCAL_KEYVAULT_PATH=keyvault" \
```
kairoaraujo, line 32: Add here:

```command
--env="RSTUF_LOCAL_KEYVAULT_PASSWORD=mypass" \
```

kairoaraujo (changes requested):
Overall it looks good, small required fixes :)

kairoaraujo (approved):

MVrachev:
I added https://github.com/MVrachev/repository-service-tuf-worker/commit/108ddc0d967bbc93d19cfde1be3766a3f2622519 and I am not sure why GitHub doesn't update this pr...

kairoaraujo (changes requested):

repository_service_tuf_worker/services/keyvault/local.py
```diff
@@ -22,87 +25,76 @@ class LocalKeyVault(IKeyVault):
...
-        self._online_key_password: Optional[str] = online_key_pass
-        self._online_key_type: Optional[str] = online_key_type
-        self._keyvault = Dynaconf(
-            envvar_prefix="LOCALKEYVAULT",
-            settings_files=[self._secrets_file],
-        )
+        self._password: str = key_pass
+        self._path: str = key_path
+        self._type: str = key_type
+        self._secrets_handler: Callable = lambda *a: self._password
```
kairoaraujo, line 43: Based on the [suggestion for the `configure`](https://github.com/vmware/repository-service-tuf-worker/pull/210/files#r1146009654), do we also need to implement it here? [I'm testing it]
```suggestion
        self._secrets_handler: Callable
        if key_pass.startswith("/run/secrets/"):
            with open(self._password) as f:
                self._secrets_handler = lambda *a: f.read().rstrip("\n")
        else:
            self._secrets_handler = lambda *a: self._password        
```

kairoaraujo (changes requested):

repository_service_tuf_worker/services/storage/local.py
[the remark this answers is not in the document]
kairoaraujo: I suggest we revert it and fix the comment r[elated in the `repository`](https://github.com/vmware/repository-service-tuf-worker/pull/210#discussion_r1146050931).
We can work on local storage service later.

docs/source/guide/Docker_README.md
```diff
@@ -6,19 +6,21 @@ repository-service-tuf-worker is part of Repository Service for TUF (RSTUF)
...
+Other requirements include:
 
-* repository-service-tuf-api
-* Compatible Borker and Result Backend Service with
+* repository-service-tuf-api service
+* Compatible broker and result backend service with
   [Celery](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/index.html).
   Recomended: [RabbitMQ](https://www.rabbitmq.com) or [Redis](https://redis.com)
+* A key that is available online. For more instructions read the
+  [Deployment documentation](https://repository-service-tuf.readthedocs.io/en/latest/guide/deployment/server.html).
```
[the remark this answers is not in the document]
kairoaraujo: I updated the suggestion above, mentioning the new feature in the CLI to generate the Key.
It would be an excellent way to go.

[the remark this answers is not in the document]
MVrachev: I am only worry that I cannot give a link to the exact place where we will document that in read-the-docs now when it's not yet merged.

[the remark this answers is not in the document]
kairoaraujo: Because of it, I added it to the repository-service-tuf-cli guide.
There can be found when it is also merged.
You can also see that I update the feature [issue](https://github.com/vmware/repository-service-tuf/issues/207), which requires the new CLI key generation feature.

repository_service_tuf_worker/services/keyvault/local.py
MVrachev: I used a different approach

```python
        if self._password.startswith("/run/secrets/"):
            # The user has stored their password using container secrets.
            with open(self._password) as f:
                secrets_password = f.read().rstrip("\n")

            self._password = secrets_password
        self._secrets_handler: Callable = lambda *a: self._password
```

repository_service_tuf_worker/services/keyvault/local.py
```diff
@@ -22,87 +25,76 @@ class LocalKeyVault(IKeyVault):
...
-            ed25519_key = encrypt_key(key.get("key"), key.get("password"))
-            key_vault_data.append(
-                {
-                    "key": ed25519_key,
-                    "filename": key["filename"].split("/")[-1],
-                    "password": key["password"],
-                }
+            priv_key_uri = f"file:{self._path}?encrypted=true"
+            return SSlibSigner.from_priv_key_uri(
+                priv_key_uri, public_key, self._secrets_handler
```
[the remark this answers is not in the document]
MVrachev: Good catch.

kairoaraujo (approved):

PR #213, opened 2023-02-27
Flexible maintainer CD Workflow

kairoaraujo:
This PR introduces a change in the `cd.yml` and adds the `publish_container.yml  Workflow.

We release our components versions using the tags `v*` The process of release has two steps:

1) Run the functional tests (FT) using the component to be released `v*` tag (from `main` branch) against the latest releases of the components.

2) If the (FT) finishes successfully, the `cd.yml` releases the new Docker Image GitHub Registry and adds it to our Releases page in Github. (Python package in case of the CLI).

If step 1 fails, the release breaks the compatibility with the current release.

The problem we are solving is how to release a new version when it breaks the compatibility.

This PR adds the workflow dispatch option for releases. The maintainers can now go to the Actions > CD > Run Workflow, hoose the `v*` tag release, and specify the version of the components it is compatible with, including dev (main branch) to run the FT tests.

Using the Workflow dispatcher process, the Maintainers should bump to a version that shows it is breaking compatibility and make it clear on the release page.

We have a (release process)[https://repository-service-tuf.readthedocs.io/en/latest/devel/release.html#release-process] and we are not changing it, only adding the Workflow that allows those actions.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -28,7 +28,6 @@ jobs:
     needs: publish-container-image
     runs-on: ubuntu-latest
     outputs:
-      worker_version: ${{ steps.worker_version.outputs.VERSION }}
       api_version: ${{ steps.api_version.outputs.VERSION }}
       cli_version: ${{ steps.cli_version.outputs.VERSION }}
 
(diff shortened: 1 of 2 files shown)


MVrachev (approved):
LGTM!

This pr is the same as the one proposed for the API: https://github.com/vmware/repository-service-tuf-api/pull/269

PR #222, opened 2023-03-09
Make tests repository and app tests independent

MVrachev:
Currently, repository.py and app.py use the worker and repository settings setup in repository_service_tuf_worker/__init__.py The way they are configured now (without a function/class directly in the module) means that when for the first time anything is imported from repository_service_tuf_worker/ it results in configuring the worker and repository settings.
As a side effect, this means that even if something is imported from repository_service_tuf_worker/ later this won't cause a new set of worker and repository settings to be set up.

This becomes a problem for our tests where in the test_repository.py and test_app.py modules we want to make sure that we have a new set of settings.
That is important because the worker and repository settings define the behavior of the repository.MetadataRepository object.

I fix that problem by just replacing the configuration of the settings to be done when a function is called instead of import.

**Note: this fix is required for #210.**

### DIFF
--- a/tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
+++ b/tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
@@ -33,7 +33,6 @@ class TestTargetsCrud:
         assert mocked_db.refresh.calls == [pretend.call("fake_db_target")]
 
     def test_read_unpublished_rolenames(self):
-
         crud.models.RSTUFTargets = pretend.stub(
             published=False, rolename="all-bins"
         )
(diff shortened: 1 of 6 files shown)


tests/unit/tuf_repository_service_worker/models/targets/test_crud.py
```diff
@@ -33,7 +33,6 @@ def test_create(self):
         assert mocked_db.refresh.calls == [pretend.call("fake_db_target")]
 
     def test_read_unpublished_rolenames(self):
-
```
MVrachev, line 36: Removing those blank lines was something `black` required.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -27,8 +26,7 @@ def test_refresh_settings_with_none_arg(self):
         test_repo.refresh_settings()
 
         assert (
-            test_repo._worker_settings.to_dict()
-            == repository.worker_settings.to_dict()
+            isinstance(test_repo._worker_settings, repository.Dynaconf) is True
```
MVrachev, line 29: We cannot check the `test_repo._worker_settings` against anything because the `worker_settings` are created internally in the `MetadataRepository` class instance by calling `get_worker_settings`.

kairoaraujo (approved):

PR #237, opened 2023-03-27
Update LocalStorage.get() to follow the interface & remove _load()

MVrachev:
Make sure that LocalStorage.get() and IStorage.get() return the same
type: Metadata[T].
Currently, the LocalStorage.get() behaves like a context manager and
returns "BufferedReader" without a good reason for that and doesn't
follow what the interface has defined as a return type - Metadata[T].

Additionally, I am removing repository._load() as we are making a mistake in passing
self.storagebackend() as an argument to "Metadata.from_file()" where it
expects an object from the type securesystemslib.StorageBackendInterface
but our self._storagebackend() doesn't implement that interface.
Finally, there are more checks made inside LocalStorage.get()
and this will allow us to be more flexible when we introduce other
storage services.

### DIFF
--- a/repository_service_tuf_worker/services/storage/local.py
+++ b/repository_service_tuf_worker/services/storage/local.py
@@ -5,14 +5,13 @@
 import glob
 import os
 import stat
-from contextlib import contextmanager
-from io import BufferedReader
 from typing import List, Optional
 
 from securesystemslib.exceptions import StorageError  # noqa
+from tuf.api.serialization import DeserializationError
 
 from repository_service_tuf_worker.interfaces import IStorage, ServiceSettings
-from repository_service_tuf_worker.repository import Timestamp
+from repository_service_tuf_worker.repository import Metadata, Timestamp
 
 
 class LocalStorage(IStorage):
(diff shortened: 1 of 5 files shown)


repository_service_tuf_worker/services/storage/local.py
```diff
@@ -62,9 +60,9 @@ def get(self, role, version=None) -> BufferedReader:
         file_object = None
         try:
             file_object = open(filename, "rb")
-            yield file_object
-        except OSError:
-            raise StorageError(f"Can't open Role '{role}'")
+            return Metadata.from_bytes(file_object)
```
kairoaraujo, line 63: It fails once it doesn't read the `file_object`:

```log
repository-service-tuf-worker_1  | Traceback (most recent call last):
repository-service-tuf-worker_1  |   File "/usr/local/lib/python3.10/site-packages/tuf/api/serialization/json.py", line 36, in deserialize
repository-service-tuf-worker_1  |     json_dict = json.loads(raw_data.decode("utf-8"))
repository-service-tuf-worker_1  | AttributeError: '_io.BufferedReader' object has no attribute 'decode'
```

```suggestion
            return Metadata.from_bytes(file_object.read())
```

kairoaraujo (changes requested):

repository_service_tuf_worker/services/storage/local.py
MVrachev: Thanks for noticing that!
I will try to test my prs better so you don't have to notice these things.

MVrachev:
I tested this code by deploying an RSTUF worker instance using these changes and the image of RSTUF API based on the changes from pr https://github.com/vmware/repository-service-tuf-api/pull/262.
The bootstrap was successful, but right now I am not able to add a new target because I cannot access the online key.
The current state of the RSTUF worker after those changes is that we have defined the configuration options for the online key, but we don't use it.
I am starting to use it in my other pr dependant on that one: #210.

kairoaraujo (approved):

PR #239, opened 2023-03-30
Update roles-simplification with latest main

kairoaraujo:
It releases the current roles-simplification using the latest main.

**DO NOT SQUASH & MERGE**


kairoaraujo:
I will open a branches merge PR instead.

PR #275, opened 2023-04-24
Repository Settings write implementation

kairoaraujo:
This commit implements two new functionalities in the Repository

* `write_repository_settings`

This method has the functionality to write settings values to the Repository settings. The repository settings uses Dynaconf configured to the Redis Server (`RSTUF_REDIS_SERVER`). This configuration is shared accross the Workers (if multiple).

This function receives a key and value and stores it to the Redis server

* `save_settings`

Uses the `write_repository_settings` to save all configuration from the Repository, which is a defined payload from the RSTUF API service. this function also requires the root TUF metadata.

settings payload:
```json
{
    "expiration": {
        "root": 365,
        "targets": 365,
        "snapshot": 1,
        "timestamp": 1,
        "bins": 1
    },
    "services": {
        "targets_base_url": "http://www.example.com/repository/",
        "number_of_delegated_bins": 4,
        "targets_online_key": True
    }
}
```

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -428,9 +428,7 @@ class MetadataRepository:
             threshold = 1
             num_of_keys = 1
             if rolename == Roles.ROOT.value.upper():
-                # The key to the root role is the name of the root file which
-                # uses consistent snapshot or in the format:
-                # <VERSION_NUMBER>.root.json
+                # get treshold and number of keys from given root metadata
                 threshold = root.signed.roles[role.value].threshold
                 num_of_keys = len(root.signatures)
 
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -15,6 +15,7 @@ import redis
 from celery.app.task import Task
 from celery.exceptions import ChordError
 from celery.result import AsyncResult, states
+from dynaconf.loaders import redis_loader
 from securesystemslib.exceptions import StorageError  # type: ignore
 from tuf.api.metadata import (  # noqa
     SPECIFICATION_VERSION,
@@ -202,6 +203,21 @@ class MetadataRepository:
 
         self._worker_settings = settings
         return settings
+
+    def write_repository_settings(self, key: str, value: Any):
+        """
+        Writes repository settings.
+
+        Repository settings uses Dynaconf and it is persistent in the Redis
+        server. It stores as a dictionary with key and value.
+
+        Args:
+            key: key name
+            value: value for the key
+        """
+        settings_data = self._settings.as_dict(env=self._settings.current_env)
+        settings_data[key] = value
+        redis_loader.write(self._settings, settings_data)
 
     def _sign(self, role: Metadata) -> None:
         """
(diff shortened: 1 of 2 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -203,6 +204,21 @@ def refresh_settings(self, worker_settings: Optional[Dynaconf] = None):
         self._worker_settings = settings
         return settings
 
+    def write_repository_settings(self, key: str, value: Any):
+        """
+        Writes repository settings.
+
+        Repository settings uses Dynaconf and it is persistent in the Redis
+        server. It stores as a dictionary with key and value.
```
MVrachev, line 212: I think this needs some additional information.
Tell me what you think of my suggestion.

``` suggestion
        Repository settings are stored in a dictionary like Dynaconf object
        and each of them has its own key and value.
        Additionally, repository settings are persisted in the Redis server
        so that they can be reused by multiple RSTUF Worker instances.
```

repository_service_tuf_worker/repository.py
```diff
@@ -203,6 +204,21 @@ def refresh_settings(self, worker_settings: Optional[Dynaconf] = None):
...
+
+        Args:
+            key: key name
+            value: value for the key
+        """
+        settings_data = self._settings.as_dict(env=self._settings.current_env)
```
MVrachev, line 218: I read the Dynaconf documentation and I am not sure what the `env` argument is used for.
Does it filter which environmental settings to return?

repository_service_tuf_worker/repository.py
```diff
@@ -203,6 +204,21 @@ def refresh_settings(self, worker_settings: Optional[Dynaconf] = None):
...
+        settings_data[key] = value
```
MVrachev, line 219: By changing the settings data don't you also need to save it back in the `Dynaconf` `self._settings` object?

repository_service_tuf_worker/repository.py
```diff
@@ -396,6 +412,46 @@ def _send_publish_targets_task(self, task_id: str):  # pragma: no cover
...
+        """
+        logging.info("Saving settings")
+        for role in Roles:
+            rolename = role.value.upper()
+            threshold = 1
+            num_of_keys = 1
+            if rolename == Roles.ROOT.value.upper():
+                # The key to the root role is the name of the root file which
+                # uses consistent snapshot or in the format:
+                # <VERSION_NUMBER>.root.json
```
MVrachev, line 431: I don't think that's true.
In our last iteration of bootstrap ceremony payload changes, we made the key to the `root` information to be just `root` and not the metadata file name anymore or using the version in any way.

repository_service_tuf_worker/repository.py
kairoaraujo: `env` allows the user to define setups for different environments.
It is used a lot by DevOps to have multiple settings in a Dynaconf configuration such as for development, quality, or production. They use it for continues deployment when using infrastructure as code.

Why do I enforce it? because we are writing in this configuration, and to ensure we are overwriting it properly, we enforce to reuse the same env it was defined.
If it is not defined, Dynaconf will use the "default"

We don't document it (maybe in the future), but in an infrastructure as a code, the user could use the same pipeline to deploy multiple workers but in different environments where they will have the specific configuration.

repository_service_tuf_worker/repository.py
kairoaraujo: The `self._settings.get_fresh()` [manages to force load from the dynaconf store](https://dynaconf.readthedocs.io/en/docs_223/reference/dynaconf.html#dynaconf.base.Settings.fresh).

I did not understand your question here—what do you mean by calling the `save` function?
The flow for save settings is as follows:
`bootstrap()` receives settings as a dict and calls `save_settings` (which prepares the keys/values) and calls `write_repository_settings()`, [which basics use the Dynaconf `redis_loader.write` to persist it on Redis](https://dynaconf.readthedocs.io/en/docs_223/reference/dynaconf.loaders.html#dynaconf.loaders.redis_loader.write).
The same will be done in the future by `metadata_rotate`, which basically will overwrite this settings.

MVrachev: After a conversation with @kairoaraujo I realized that `redis_loader` below is an API call from Dynaconf and given that the `self_settings` Dynaconf object uses `Redis` as an internal storage it will make sure that it writes this new setting in the `Redis` server and at `Dynaconf` object as well.

MVrachev (approved):

PR #278, opened 2023-04-25
Update container docs about secrets

kairoaraujo:
Add the mention about the secrets standard adopted by RSTUF. RSTUF requires user to mount the secrets volume in `/run/secrets` path.

### DIFF
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -130,8 +130,8 @@ Available types:
   - Required variables:
     - ``RSTUF_LOCAL_KEYVAULT_PASSWORD``
       - password used to load the online key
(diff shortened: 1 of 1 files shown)


docs/source/guide/Docker_README.md
```diff
@@ -130,6 +130,9 @@ Available types:
   - Required variables:
     - ``RSTUF_LOCAL_KEYVAULT_PASSWORD``
       - password used to load the online key
+      - This environment variable supports container secrets when the volume is
+        added to `/run/secrets` path.
```
MVrachev, line 134: Isn't more correct to say:
```suggestion
      - This environment variable supports container secrets when the `/run/secrets`
        volume is added to the path.
```

MVrachev (approved):

PR #280, opened 2023-05-01
Improve the SQL worker container configuration

### LINKED ISSUE #231: Security: Give possibility to use container secrets for `RSTUF_SQL_SERVER` environment variable

The current container environment variable [`RSTUF_SQL_SERVER`](https://github.com/vmware/repository-service-tuf-worker/blob/main/docs/source/guide/Docker_README.md#required-rstuf_sql_server) currently uses a text string.

**Should be interesting to enable to use secrets?**

Example:

```shell
    $ docker swarm init
    $ printf "postgresql://postgres:secret@postgres:5432" | docker secret create RSTUF_SQL_SERVER -
```

Then define it as secrete, for example, in docker-compose.yml

```yml
      RSTUF_SQL_SERVER: /run/secrets/RSTUF_SQL_SERVER

```

### References

_No response_

### Code of Conduct

kairoaraujo:
This commit improves and gives users more flexibility to configure the SQL server on RSTUF Worker.

The user can now use secrets instead of plain text to configure the SQL server password.

The user is still able to use the previous configuration.

Closes #231

### DIFF
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -73,13 +73,30 @@ Example: `redis://redis`
 
 #### (Required) `RSTUF_SQL_SERVER`
 
-SQL server address.
+RSTUF requires [PostgreSQL](https://www.postgresql.org).
 
-The SQL Server must to be compatible with
-[SQLAlchemy](https://www.sqlalchemy.org). RSTUF recomends
-[PostgreSQL](https://www.postgresql.org).
+Example: `postgres:secret@postgres:5432`
 
-Example: `postgresql://postgres:secret@postgres:5432`
+* Optional variables:
+
+  * `RSTUF_SQL_USER` optional information about the user name
+
+    If using this optional variable:
+    - Do not include user in the `RSTUF_SQL_SERVER`.
+    - The `RSTUF_SQL_PASSWORD` becomes required
+
+  * `RSTUF_SQL_PASSWORD` use this variable to inform the password separated.
+    - Do not include the password in the `RSTUF_SQL_SERVER`
+    - This environment variable supports container secrets when the `/run/secrets`
+      volume is added to the path.
+
+  Example:
+  ```
+  RSTUF_SQL_SERVER=sqlserver:5432
+  RSTUF_SQL_USER=postgres
+  RSTUF_SQL_PASSWORD=/run/secrets/POSTGRES_PASSWORD
+  ```
+
 
 #### (Optional) `RSTUF_REDIS_SERVER_PORT`
 
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -82,10 +82,10 @@ Example: `postgres:secret@postgres:5432`
   * `RSTUF_SQL_USER` optional information about the user name
 
     If using this optional variable:
-    - Do not include user in the `RSTUF_SQL_SERVER`.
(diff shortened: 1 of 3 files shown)


docs/source/guide/Docker_README.md
```diff
@@ -73,13 +73,30 @@ Example: `redis://redis`
 
 #### (Required) `RSTUF_SQL_SERVER`
 
-SQL server address.
+RSTUF requires [PostgreSQL](https://www.postgresql.org).
 
-The SQL Server must to be compatible with
-[SQLAlchemy](https://www.sqlalchemy.org). RSTUF recomends
-[PostgreSQL](https://www.postgresql.org).
```
MVrachev, line 80: By removing those and replacing them with line 76 we insist that we accept only `PostgresSQL`, right?

docs/source/guide/Docker_README.md
```diff
@@ -73,13 +73,30 @@ Example: `redis://redis`
...
+Example: `postgres:secret@postgres:5432`
+
+* Optional variables:
+
+  * `RSTUF_SQL_USER` optional information about the user name
+
+    If using this optional variable:
+    - Do not include user in the `RSTUF_SQL_SERVER`.
```
MVrachev, line 85:
```suggestion
    - Do not include the user in the `RSTUF_SQL_SERVER`.
```

docs/source/guide/Docker_README.md
```diff
@@ -73,13 +73,30 @@ Example: `redis://redis`
...
+    - The `RSTUF_SQL_PASSWORD` becomes required
+
+  * `RSTUF_SQL_PASSWORD` use this variable to inform the password separated.
```
MVrachev, line 88:
```suggestion
  * `RSTUF_SQL_PASSWORD` use this variable to provide the password separately.
```

repository_service_tuf_worker/repository.py
```diff
@@ -106,11 +106,31 @@ def refresh_settings(self, worker_settings: Optional[Dynaconf] = None):
...
-
+        #
         # SQL
-        settings.SQL = rstuf_db(self._worker_settings.SQL_SERVER)
+        #
+        sql_server_url = self._worker_settings.get("SQL_SERVER")
+        # clean 'postgresql://' if present
+        sql_server = sql_server_url.replace("postgresql://", "")
+        if sql_user := self._worker_settings.get("SQL_USER"):
+            if self._worker_settings.SQL_PASSWORD.startswith("/run/secrets"):
```
MVrachev, line 116: If we miss `SQL_PASSWORD` here do we want to log the error better or just leave `AttributeError` as our error of choice?
Maybe if we raise an error we can say: `SQL_PASSWORD required because SQL_USER was used`.
But maybe the error `Settings' object has no attribute 'SQL_PASSWORD` is good enough.

MVrachev (commented):
Good changes and great tests!
I have only clarifying questions.

docs/source/guide/Docker_README.md
kairoaraujo: Yes, correct.
Let's start by suggesting a database we have tested and used.

MVrachev (approved):

repository_service_tuf_worker/repository.py
kairoaraujo: Filed issue:
- https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/284

PR #311, opened 2023-06-15
Refactoring metadata update

### LINKED ISSUE #302: Refactoring to `metadata_update` and add validation on the new metadata
https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/e57d15054aa91270d2006676fe06bde0c031d8bf/repository_service_tuf_worker/repository.py#L980

Rename the function to `metadata_update`

To avoid breaking compatibility, keep the old with a deprecation warning

Add a check if the metadata is trusted update before accepting

See https://github.com/theupdateframework/python-tuf/blob/61d29b32e9164beb11b2e9b8140b2ecd34b2d035/tuf/ngclient/updater.py#L323

kairoaraujo:
It aligns the process name with the function name of the other
components (API and CLI).

Renames the function from `metadata_rotation` to `metadata_update`.

Add the `metadata_rotation` function, which calls the
`metadata_update` but also gives a warning and warning logs.
It gives compatibility for users using the RSTUF API v2.3.0b1 with
RSTUF Worker until version >= v0.4.0b1 > v1.0.0.

Fix the unit tests and add an extra test to check the deprecation warning.

Add support to multiple metadata rotation types 
Even now, the RSTUF process uses Online Key only for Root Metadata; this
PR improves the Metadata Update process to support the future
multiple Metadata types.

This also makes more clear with the helper function, where all the Root
Metadata process.

Add/extend root check in a helper function 
Add some trust checks in the root metadata. It is based on python-tuf
checks.

Closes #302

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -6,6 +6,7 @@ import enum
 import importlib
 import logging
 import time
+import warnings
 from dataclasses import asdict, dataclass
 from datetime import datetime, timedelta
 from math import log
@@ -977,7 +978,7 @@ class MetadataRepository:
 
         return True
 
-    def metadata_rotation(
+    def metadata_update(
         self,
         payload: Dict[Literal["metadata"], Dict[Literal[Root.type], Any]],
         update_state: Optional[
@@ -987,7 +988,7 @@ class MetadataRepository:
         """
         Rotate TUF metadata.
 
-        Adds support for metadata rotation signed by offline root keys.
+        Adds support for metadata update signed by offline root keys.
 
         Args:
             payload: contains new metadata
@@ -1003,9 +1004,7 @@ class MetadataRepository:
             raise KeyError("No 'root' in the 'metadata' payload.")
 
         if self._settings.get_fresh("BOOTSTRAP") is None:
-            raise RuntimeError(
-                "metadata rotation requires RSTUF with bootstrap"
-            )
+            raise RuntimeError("metadata update requires RSTUF with bootstrap")
 
         new_root: Metadata[Root] = Metadata.from_dict(root_dict)
         current_root: Metadata[Root] = self._storage_backend.get(Root.type)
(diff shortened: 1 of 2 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -1057,9 +1064,59 @@ def metadata_rotation(
...
+
+    def metadata_rotation(
+        self,
+        payload: Dict[Literal["metadata"], Dict[Literal[Root.type], Any]],
+        update_state: Optional[
+            Task.update_state
+        ] = None,  # It is required (see: app.py)
+    ) -> Dict[str, Any]:
+        deprecation_message = (
+            "Function `metadata_rotation` will be deprecated on version "
```
lukpueh, line 1116: nit: IIUC you're deprecating it now and removing it in 1.0.0?

repository_service_tuf_worker/repository.py
```diff
@@ -977,43 +978,49 @@ def bump_online_roles(self, force: Optional[bool] = False) -> bool:
...
+
+        # Verify the new root version
+        if new_root.signed.version != current_root.signed.version + 1:
+            raise BadVersionNumberError(
+                f"Expected root version {current_root.signed.version + 1}"
+                f" instead got version {new_root.signed.version}"
+            )
+
+        # Verify the Type
+        if new_root.signed.type != Root.type:
```
lukpueh, line 999: Any reason to fail so late? You could check this right after Metadata.from_dict, before fetching current root from storage or verifying any signatures.

lukpueh (approved):
LGTM. My comments are definitely no blockers.

repository_service_tuf_worker/repository.py
kairoaraujo: Yes. I will fix the sentence.

repository_service_tuf_worker/repository.py
kairoaraujo: I think the reason was to put all
checks together. But we can check it once we receive the payload and load the new root.

I don't have a preference, to be honest.

kairoaraujo: I didn't move it out of the function but added it as a first check.

lukpueh (approved):
🚀

Issue #315, opened 2023-07-06
Enhancement: Stop raising exception for tasks when we can handle and gracefully finish the task 


kairoaraujo:
### What do you want to share with us?

We are mishandling the tasks in many RSTUF Worker tasks.
We should stop raising exceptions for tasks when we can handle and gracefully finish the task.
I noticed it when implementing the [Distributed Async Singing in the Worker](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/313).

#### Introduction

By design RSTUF is asynchronous, and we use a message queue to publish tasks (from RSTUF API).
[The RSTUF Worker consumes these tasks, computes them, and stores the result backend](https://repository-service-tuf.readthedocs.io/projects/rstuf-worker/en/latest/devel/index.html).

To implement it, we use Celery.  The Celery has some of the [tasks states](https://docs.celeryq.dev/en/stable/userguide/tasks.html#built-in-states).

### The problem

We use the exception to handle the task result as a task state.

We shouldn't have the task as `FAILURE`, but `SUCCESS` and add the result to the task.
A task `SUCCESS` means that the task finished with `SUCCESS`, but the result will say what the result of this task was.
The `FAILURE` should be used when the task can't finish for some reason and we cannot handle it.

Task state is the computing of the task
Task result is the result of the task computation

### Solution

We should handle the errors and back the status to the user, not making the task fail once our implementation handled it.

Example 1:

We just failed the successful task, but the result is a bootstrap failure.

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/833c5056aa0d09ab8cb0d2cb6600e3ff24baa9d7/repository_service_tuf_worker/repository.py#L527-L528

We shouldn't raise an exception here, but we should return the [`ResultDetails`](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/833c5056aa0d09ab8cb0d2cb6600e3ff24baa9d7/repository_service_tuf_worker/repository.py#L75)

```python
result = ResultDetails(
    status="bootstrap failed",
    details={
        "bootstrap": False,
        "message": "Missing 'settings'",
    },
    last_update=datetime.now(),
)
```

Example 2:

During a Metadata Update, executing the `_root_metadata_update`, we call https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/833c5056aa0d09ab8cb0d2cb6600e3ff24baa9d7/repository_service_tuf_worker/repository.py#L1024

[...]

PR #337, opened 2023-07-26
Add task "update_settings" in Repository

### LINKED ISSUE #297: Task: add task "update_settings" in Repository
### What is the task about?

This endpoint is suggested in the RSTUF API issue: https://github.com/repository-service-tuf/repository-service-tuf-api/issues/350
We want to add support for it.

### Parent feature

- https://github.com/repository-service-tuf/repository-service-tuf/issues/347

### References

_No response_

### Code of Conduct

MVrachev:
Add PUT /api/config endpoint to update settings

The API implementation is here: https://github.com/repository-service-tuf/repository-service-tuf-api/pull/387

Close  #297.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -57,6 +57,10 @@ class Roles(enum.Enum):
     TIMESTAMP = Timestamp.type
     BINS = "bins"
 
+    @staticmethod
+    def online_roles() -> List[str]:
+        return [Targets.type, Snapshot.type, Timestamp.type, "bins"]
+
 
 ALL_REPOSITORY_ROLES_NAMES = [rolename.value for rolename in Roles]
 OFFLINE_KEYS = {
@@ -242,6 +246,7 @@ class MetadataRepository:
             key: key name
             value: value for the key
         """
+        logging.info(f"Saving {key} with value: {value}")
         settings_data = self._settings.as_dict(env=self._settings.current_env)
         settings_data[key] = value
         redis_loader.write(self._settings, settings_data)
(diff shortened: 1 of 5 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -628,6 +633,34 @@ def bootstrap(
...
+        self,
+        payload: Dict[str, Any],
+        update_state: Optional[
+            Task.update_state
+        ] = None,  # It is required (see: app.py)
+    ):
+        """Update repository settings with the new settings."""
+        tuf_settings: Dict[str, Any] = payload.get("settings")
+        if tuf_settings is None:
+            raise KeyError("No 'settings' in the payload")
```
kairoaraujo, line 646: nit: We should [stop raising exceptions to finish the task](https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/315).
We should return the task with the details that `update_repository` as `False`

repository_service_tuf_worker/repository.py
```diff
@@ -628,6 +633,34 @@ def bootstrap(
...
+
+        if tuf_settings.get("expiration") is None:
+            raise KeyError("No 'expiration' in the payload")
+
+        if len(tuf_settings["expiration"]) < 1:
+            raise KeyError("No role provided for expiration policy change")
+
+        logging.info("Updating settings")
+        online_roles = Roles.online_roles()
+        for role in tuf_settings["expiration"]:
```
kairoaraujo, line 656: Following the implementation, it will case that some expiration can be updated and others roles not if it receives a payload with an invalid role.
It brings a problem to the user because the task will fail in the middle with partial change.

IMO we could apply the same behavior we do deleting targets (artifacts)
When we return the task, we say which ones are changed and which are invalid.

repository_service_tuf_worker/repository.py
```diff
@@ -628,6 +633,34 @@ def bootstrap(
...
+            if role not in online_roles:
+                raise KeyError(f"Role {role} is not a valid dictionary key")
+
+            self.write_repository_settings(
+                f"{role.upper()}_EXPIRATION", tuf_settings["expiration"][role]
+            )
+
```
kairoaraujo, line 663: We don't return the task result here in the end.
To the user checking the task status will be `None`

kairoaraujo (changes requested):

kairoaraujo (commented):
Maybe we can simplify using two words.

repository_service_tuf_worker/repository.py
```diff
@@ -628,6 +633,76 @@ def bootstrap(
...
+            return ResultDetails(
+                status="update settings failed",
+                details={
+                    "update_settings": False,
+                    "message": "No 'expiration' in the payload",
+                    "updated_roles": [],
+                    "non_valid_roles": [],
+                },
+                last_update=datetime.now(),
+            )
```
kairoaraujo, line 667: For the above three returns, I don't think we env need the updated and invalid roles in the details.

repository_service_tuf_worker/repository.py
```diff
@@ -628,6 +633,72 @@ def bootstrap(
...
         return asdict(result)
 
+    def update_settings(
+        self,
+        payload: Dict[str, Any],
+        update_state: Optional[
+            Task.update_state
+        ] = None,  # It is required (see: app.py)
+    ) -> ResultDetails:
+        """Update repository settings with the new settings."""
```
kairoaraujo, line 643:
```suggestion
        """
        Update repository settings with the new settings.

        It supports only the expiration settings for online roles.
        Expiration parameters reference:
        https://repository-service-tuf.readthedocs.io/en/stable/devel/design.html#rstuf-repository-settings-configuration  # noqa
        """
```

kairoaraujo (approved):
nit: Add on docstring the information that it supports only expiration and a reference for developers.

PR #350, opened 2023-08-10
Fix RSTUF Worker container image release tag

kairoaraujo:
Fix the RSTUF Worker container image release tag.

The tag `latest` wasn't being added to the releases confusing the users running old versions.


rdimitrov (approved):
LGTM! 👍

Issue #336, opened 2023-07-25
Task: Implement Metadata Update Distributed Async Signature feature


kairoaraujo:
### What is the task about?

This task is about the Worker running tasks for Signing a TUF metadata.

It requires:
- #300, which implements the base `sign_metadata`

Required changes:

* Metadata Update
  - It does need to check the signing state (threshold) like in Bootstrap and update TUF Settings (Redis `<ROLENAME>_SIGNING`) and persist the metadata.
  - It is important to validate the signatures threshold of the previous and new metadata to have consistency. [Specification reference.](https://theupdateframework.github.io/specification/latest/#key-management-and-migration)
  
**Note:** RSTUF currently has only one Role with offline key(s), but this process might be helpful in the future to support other offline roles or even support delegate roles using offline keys.

### Parent feature

- https://github.com/repository-service-tuf/repository-service-tuf/issues/327

### References

- #300
- https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/313#discussion_r1269408892

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MVrachev:
We should be able to close this issue, right @lukpueh and @kairoaraujo?

PR #355, opened 2023-08-14
Support 'metadata update' in 'sign_metadata' task

lukpueh:
related to #336

Implement support for distributed asynchronous root metadata signging in the course of a "metadata update" event.

Other than the already supported "bootstrap" signing event, signatures added to root during "metadata update" must validate with keys from trusted OR new root, and meet the signature threshold of trusted AND new root.

*Related changes:*
- Refactor `_validate_{signature, threshold}` helpers to accept an optional delegator (e.g. trusted root).
- Add `_sign_result` helper to return a "sign metadata"-specific task result.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -1146,6 +1146,16 @@ class MetadataRepository:
                 },
             )
 
+        self._root_metadata_update_finalize(current_root, new_root)
+        return self._task_result(
+            TaskName.METADATA_UPDATE,
+            True,
+            {"message": "Metadata Update Processed", "role": Root.type},
+        )
+
+    def _root_metadata_update_finalize(
+        self, current_root: Metadata[Root], new_root: Metadata[Root]
+    ) -> None:
         # We always persist the new root metadata, but we cannot persist
         # without verifying if the online key is rotated to avoid a mismatch
         # with the rest of the roles using the online key.
@@ -1184,12 +1194,6 @@ class MetadataRepository:
                         "RSTUF: Task exceed `LOCK_TIMEOUT` "
                         f"({self._timeout} seconds)"
                     )
-
-        return self._task_result(
-            TaskName.METADATA_UPDATE,
-            True,
-            {"message": "Metadata Update Processed", "role": Root.type},
-        )
 
     def metadata_update(
         self,
@@ -1409,16 +1413,8 @@ class MetadataRepository:
                 msg = f"Root v{root.signed.version} is pending signatures"
                 return _result(True, update=msg)
 
-            # TODO: Refactor `_root_metadata_update` to de-duplicate validation
-            # and messaging. At this point, we know that root is valid and
-            # there can be only one message. (remove assert after refactor!)
-            result = self._root_metadata_update(root)
-            assert result == {  # nosec
-                "message": "Metadata Update Processed",
-                "role": "root",
-            }
-
-            # Update successful, root persisted -> finalize event...
+            # Threshold reached -> finalize event
+            self._root_metadata_update_finalize(trusted_root, root)
             self.write_repository_settings("ROOT_SIGNING", None)
             return _result(True, update="Metadata update finished")
 
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -1262,20 +1262,25 @@ class MetadataRepository:
         return self.metadata_update(payload, update_state)
 
     @staticmethod
-    def _validate_signature(metadata: Metadata, signature: Signature) -> bool:
+    def _validate_signature(
+        metadata: Metadata,
+        signature: Signature,
+        delegator: Optional[Metadata] = None,
+    ) -> bool:
         """
-        Validate  signature over metadata using appropriate delegator.
-
-        NOTE: In "metadata update" signing event, the public key and
-        authorization info is retrieved from "trusted root"
+        Validate signature over metadata using appropriate delegator.
+        If no delegator is passed, the metadata itself is used as delegator.
 
         """
+        if delegator is None:
+            delegator = metadata
+
         keyid = signature.keyid
-        if keyid not in metadata.signed.roles[Root.type].keyids:
+        if keyid not in delegator.signed.roles[Root.type].keyids:
             logging.info(f"signature '{keyid}' not authorized")
             return False
 
-        key = metadata.signed.keys.get(signature.keyid)
+        key = delegator.signed.keys.get(signature.keyid)
         if not key:
             logging.info(f"no key for signature '{keyid}'")
             return False
@@ -1292,25 +1297,19 @@ class MetadataRepository:
         return True
 
     @staticmethod
-    def _validate_threshold(metadata: Metadata) -> bool:
+    def _validate_threshold(
+        metadata: Metadata, delegator: Optional[Metadata] = None
+    ) -> bool:
         """
         Validate signature threshold using appropriate delegator(s).
+        If no delegator is passed, the metadata itself is used as delegator.
 
-        NOTE: In "metadata update" signing event, the threshold for:
-        -  root is validated with the passed metadata AND the trusted root;
-        -  top-level targets is validated with the trusted root;
-        -  delegated targets is validated with the delegating targets;
-        as delegator.
         """
+        if delegator is None:
+            delegator = metadata
 
         try:
-            # TODO: `verify_delegate` does not tell us if there are any
-            # superfluous valid or invalid signatures. Is this something we
-            # want to know, e.g. to detect mistakes? To detect superfluous
-            # signatures if verify_delegate succeeds, would be easy: `assert
-            # len(signatures) == threshold`. Anything, else would require a
-            # custom `verify_delegate` function.
-            metadata.verify_delegate(Root.type, metadata)
+            delegator.verify_delegate(Root.type, metadata)
 
         except UnsignedMetadataError as e:
             logging.info(e)
@@ -1327,100 +1326,87 @@ class MetadataRepository:
     ) -> Dict[str, Any]:
         """Add signature to metadata for pending signing event.
 
-        Add signature (from payload) to cached role metadata (from settings)
-        for the role that matches the passed rolename (from payload), if a
-        signing event exists for that role, and the signature is valid.
+        Add signature (from payload) to cached root metadata (from settings),
+        if a signing event exists, and the signature is valid.
 
-        If the signature threshold is reached, the signing event is finalized.
+        Signing event types are 'bootstrap' or 'metadata update'.
 
-        ** Signing event types (and details) **
-
-        BOOTSTRAP: Only root metadata can be updated in this event. To verify
-        the passed signature, the keys and threshold are read from the root
-        metadata to be updated itself. If the threshold is reached, the
-        bootstrap process is finalized.
-
-        METADATA UPDATE: Root, targets or delegated targets metadata can be
-        updated in this event. Depending on the metadata type, the authorized
-        public keys and threshold are read from different delegating metadata.
+        If the signature threshold is reached, the signing event is finalized,
+        otherwise it remains in pending state.
         """
-        rolename = payload["role"]
-        signature_dict = payload["signature"]
 
-        # Assert pending signing event
-        metadata_dict = self._settings.get_fresh(f"{rolename.upper()}_SIGNING")
+        def _result(status, error=None, bootstrap=None, update=None):
+            details = {}
+            if status:
+                details["message"] = "Signature Processed"
+            else:
+                details["message"] = "Signature Failed"
+            if error:
+                details["error"] = error
+            if bootstrap:
+                details["bootstrap"] = bootstrap
+            if update:
+                details["update"] = update
+            return self._task_result(TaskName.SIGN_METADATA, status, details)
+
+        signature = Signature.from_dict(payload["signature"])
+
+        # Assert pending signing event exists
+        metadata_dict = self._settings.get_fresh("ROOT_SIGNING")
         if metadata_dict is None:
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": f"No signatures pending for {rolename}",
-                },
-            )
+            msg = "No signatures pending for root"
+            return _result(False, error=msg)
 
-        # Assert signing event type (currently bootstrap only)
-        # TODO: repository-service-tuf/repository-service-tuf-worker#336
-        bootstrap_state = self._settings.get_fresh("BOOTSTRAP")
-        if "signing" not in bootstrap_state:
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": "No bootstrap available for signing",
-                },
-            )
-
-        # Assert metadata type is allowed for signing event
+        # Assert metadata type is root
         root = Metadata.from_dict(metadata_dict)
         if not isinstance(root.signed, Root):
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": f"Role {rolename} has wrong type",
-                },
-            )
+            msg = f"Expected 'root', got '{root.signed.type}'"
+            return _result(False, error=msg)
 
-        # Assert passed signature is valid for metadata
-        signature = Signature.from_dict(signature_dict)
-        if not self._validate_signature(root, signature):
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": "Invalid signature",
-                },
-            )
+        # If it isn't a "bootstrap" signing event, it must be "update metadata"
+        bootstrap_state = self._settings.get_fresh("BOOTSTRAP")
+        if "signing" in bootstrap_state:
+            # Signature and threshold of initial root can only self-validate,
+            # there is no "trusted root" at bootstrap time yet.
+            if not self._validate_signature(root, signature):
+                return _result(False, error="Invalid signature")
 
-        # Check threshold with new signature included
-        root.signatures[signature.keyid] = signature
-        if not self._validate_threshold(root):
-            self.write_repository_settings("ROOT_SIGNING", root.to_dict())
-            root_version = root.signed.version
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                True,
-                {
-                    "message": "Signature Processed",
-                    "bootstrap": f"Root v{root_version} is pending signatures",
-                },
-            )
+            root.signatures[signature.keyid] = signature
+            if not self._validate_threshold(root):
+                self.write_repository_settings("ROOT_SIGNING", root.to_dict())
+                msg = f"Root v{root.signed.version} is pending signatures"
+                return _result(True, bootstrap=msg)
 
-        # Finalize bootstrap
-        bootstrap_task_id = bootstrap_state.split("signing-")[1]
-        self._bootstrap_finalize(root, bootstrap_task_id)
-        return self._task_result(
-            TaskName.SIGN_METADATA,
-            True,
-            {
-                "message": "Signature Processed",
-                "bootstrap": "Bootstrap Finished",
-            },
-        )
+            bootstrap_task_id = bootstrap_state.split("signing-")[1]
+            self._bootstrap_finalize(root, bootstrap_task_id)
+            return _result(True, bootstrap="Bootstrap Finished")
+
+        else:
+            # Also consult with "trusted root" when updating a new root
+            # - Signature must validate with trusted OR new root
+            # - Threshold must validate with trusted AND new root
+            trusted_root = self._storage_backend.get("root")
+            trusted_signature = self._validate_signature(
+                root, signature, trusted_root
+            )
+            new_signature = self._validate_signature(root, signature)
+
+            if not (trusted_signature or new_signature):
+                return _result(False, error="Invalid signature")
+
+            root.signatures[signature.keyid] = signature
+            trusted_threshold = self._validate_threshold(root, trusted_root)
+            new_threshold = self._validate_threshold(root)
+            if not (trusted_threshold and new_threshold):
+                self.write_repository_settings("ROOT_SIGNING", root.to_dict())
+                msg = f"Root v{root.signed.version} is pending signatures"
+                return _result(True, update=msg)
+
+            # TODO: Refactor `_root_metadata_update` to de-duplicate validation
+            self._root_metadata_update(root)
+            # Update successful, root persisted -> finalize event...
+            self.write_repository_settings("ROOT_SIGNING", None)
+            return _result(True, update="Metadata update finished")
 
     def delete_sign_metadata(
         self,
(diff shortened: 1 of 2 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -1357,92 +1373,68 @@ def sign_metadata(
...
+
+            root.signatures[signature.keyid] = signature
+            trusted_threshold = self._validate_threshold(root, trusted_root)
+            new_threshold = self._validate_threshold(root)
+            if not (trusted_threshold and new_threshold):
+                self.write_repository_settings("ROOT_SIGNING", root.to_dict())
+                msg = f"Root v{root.signed.version} is pending signatures"
+                return self._sign_result(msg, True)
+
+            # TODO: Refactor `_root_metadata_update` to de-duplicate validation
```
lukpueh, line 1436: Let me know, if I should refactor `_root_metadata_update` in this PR or in a follow-up. I think it's correct as is, but performs some unnecessary cycles.

lukpueh:
Need to fix tests (some expected messages changed) and add new ones for the new feature. Otherwise this should be good for review.

repository_service_tuf_worker/repository.py
kairoaraujo: You made a minimalist change in this PR, great work.
So, we can accommodate the refactoring here, but it is up to you.

repository_service_tuf_worker/repository.py
```diff
@@ -1421,97 +1420,84 @@ def sign_metadata(
...
+            if status:
+                details["message"] = "Signature Processed"
+            else:
+                details["message"] = "Signature Failed"
+            if error:
+                details["error"] = error
+            if bootstrap:
+                details["bootstrap"] = bootstrap
+            if update:
+                details["update"] = update
```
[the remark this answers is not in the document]
kairoaraujo: We call the task "Metadata Update", so "update" or "metadata_update" is ok IMO.
I thought in the future to identify as "Root" or "Targets", but it doesn't give much as it could be a delegated target as well.
Maybe the role type, role name, and version? Because after the request, the user has only the task id, and retrieving this task results with more details could be helpful. WDYT?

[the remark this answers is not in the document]
lukpueh: > Maybe the role type, role name, and version? Because after the request, the user has only the task id, and retrieving this task results with more details could be helpful. WDYT?

role name and version should be good. Should I add this here, or in a follow-up?

repository_service_tuf_worker/repository.py
lukpueh: I suggest to do it in a follow-up. :) I can ticketize.

repository_service_tuf_worker/repository.py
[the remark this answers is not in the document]
kairoaraujo: btw, I think is a "good first issue" too

lukpueh:
This is ready for another round of review. I adopted updates in existing tests and added new tests for:
- invalid signature (different combinations)
- valid signature and invalid threshold (different combinations)
- valid signature and valid threshold -> finalise metadata update

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
+            details = {}
+            if status:
+                details["message"] = "Signature Processed"
+            else:
+                details["message"] = "Signature Failed"
+            if error:
+                details["error"] = error
+            if bootstrap:
+                details["bootstrap"] = bootstrap
+            if update:
```
MVrachev, line 1446: You wouldn't call `bootstrap` and `update` and the same time.
So maybe it's better to remove a possibility of a mistake, make it:
```suggestion
            if bootstrap:
                details["bootstrap"] = bootstrap
            elif update:
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
+                details["update"] = update
+            return self._task_result(TaskName.SIGN_METADATA, status, details)
+
+        signature = Signature.from_dict(payload["signature"])
+
+        # Assert pending signing event exists
+        metadata_dict = self._settings.get_fresh("ROOT_SIGNING")
```
MVrachev, line 1453: We do send a role in the payload when calling `POST /api/v1/metadata/sign`, so I think we should use it here.
Also, it's not `ROLE_SIGNING` it's `SIGNING_ROLE`:
https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#rstuf-repository-settings-configuration
So, I suggest you do:
```suggestion
        metadata_dict = self._settings.get_fresh(f"SIGNING_{payload['role']}")
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
         if metadata_dict is None:
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": f"No signatures pending for {rolename}",
-                },
-            )
+            msg = "No signatures pending for root"
```
MVrachev, line 1455:
```suggestion
            msg = f"No signatures pending for {payload['role']}"
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": "No bootstrap available for signing",
-                },
-            )
-
-        # Assert metadata type is allowed for signing event
+        # Assert metadata type is root
```
MVrachev, line 1458:
```suggestion
        # Assert metadata type is payload['role']
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
         if not isinstance(root.signed, Root):
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": f"Role {rolename} has wrong type",
-                },
-            )
+            msg = f"Expected 'root', got '{root.signed.type}'"
```
MVrachev, line 1461:
```suggestion
            msg = f"Expected {payload['role']}, got '{role.signed.type}'"
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
-                False,
-                {
-                    "message": "Signature Failed",
-                    "error": "No bootstrap available for signing",
-                },
-            )
-
-        # Assert metadata type is allowed for signing event
+        # Assert metadata type is root
         root = Metadata.from_dict(metadata_dict)
```
MVrachev, line 1459:
```suggestion
        role = Metadata.from_dict(metadata_dict)
```

repository_service_tuf_worker/repository.py
```diff
@@ -1390,25 +1395,19 @@ def _validate_signature(metadata: Metadata, signature: Signature) -> bool:
         return True
 
     @staticmethod
-    def _validate_threshold(metadata: Metadata) -> bool:
+    def _validate_threshold(
+        metadata: Metadata, delegator: Optional[Metadata] = None
+    ) -> bool:
         """
         Validate signature threshold using appropriate delegator(s).
+        If no delegator is passed, the metadata itself is used as delegator.
```
MVrachev, line 1404:
```suggestion
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
+        bootstrap_state = self._settings.get_fresh("BOOTSTRAP")
+        if "signing" in bootstrap_state:
+            # Signature and threshold of initial root can only self-validate,
+            # there is no "trusted root" at bootstrap time yet.
+            if not self._validate_signature(root, signature):
+                return _result(False, error="Invalid signature")
+
+            root.signatures[signature.keyid] = signature
+            if not self._validate_threshold(root):
+                self.write_repository_settings("ROOT_SIGNING", root.to_dict())
```
MVrachev, line 1474: Again, it's not `ROLE_SIGNING` it's `SIGNING_ROLE`:
https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#rstuf-repository-settings-configuration
```suggestion
                self.write_repository_settings("SIGNING_ROOT", root.to_dict())
```

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
+                msg = f"Root v{root.signed.version} is pending signatures"
```
MVrachev, line 1475: Do we want to say how many signatures?

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
-            root_version = root.signed.version
-            return self._task_result(
-                TaskName.SIGN_METADATA,
-                True,
-                {
-                    "message": "Signature Processed",
-                    "bootstrap": f"Root v{root_version} is pending signatures",
-                },
+        else:
+            # Also consult with "trusted root" when updating a new root
```
MVrachev, line 1483: Again there could be another role that we sign.
Not sure we should stick to `root` everywhere in this code.

repository_service_tuf_worker/repository.py
```diff
@@ -1425,97 +1424,84 @@ def sign_metadata(
...
+            # - Signature must validate with trusted OR new root
```
MVrachev, line 1484: It took me a little time to get why the signature must validate with trusted OR new root.
Maybe it's worth documenting it with more details adding something like `Signature could be created either from a new key in the new root or a key existing in trusted root as well`.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -3311,3 +3267,124 @@ def fake_get_fresh(key):
         assert test_repo.write_repository_settings.calls == [
             pretend.call("ROOT_SIGNING", "fake_metadata")
         ]
+
+    @pytest.mark.parametrize(
```
MVrachev, line 3271: We had a discussion with @kairoaraujo and @KAUTH regarding table testing done in this way and we are all with mixed feelings.
The biggest issues I see with them:
1. it's harder to debug which test case fails when you see only `test_sign_metadata__update` with a specific assert that not always point you to the correct use case.
2. it could be a little complicated for new users.

What do you think @kairoaraujo and @KAUTH ?

MVrachev (changes requested):
I saw that `ROLE_SIGNING` is used in many places where we have documented `SIGNING_ROLE`: https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#rstuf-repository-settings-configuration
Do we want to update the docs or update our code?
@lukpueh @kairoaraujo

repository_service_tuf_worker/repository.py
lukpueh: The `elif` alone does not remove the possibility of mistake. I'd have to add a `raise` on `else`, which seems overkill for a nested function, whose calls I control and test. But yeah, the elif removes an unnecessary evaluation. I can change this. 👍

repository_service_tuf_worker/repository.py
lukpueh: > We do send a role in the payload when calling POST /api/v1/metadata/sign, so I think we should use it here.

You could easily argue the other way around. :) I think requiring role name in POST is premature. 
We don't know yet, how the API will evolve, and we can easily add an optional role name parameter later (root remains default), without breaking backwards compatibility. Until that it is confusing to require role name if it has to be root.

> Also, it's not ROLE_SIGNING it's SIGNING_ROLE:
https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#rstuf-repository-settings-configuration

Good catch! This was already discussed in #313, which had a mix of both. We opted for `<ROLE NAME>_SIGNING` there, which is what I continued here, but must have missed the docs in the umbrella repo. I'm fine to re-write here, or change the docs. The latter seems like a smaller change.

repository_service_tuf_worker/repository.py
lukpueh: That would be great, but it's a bit tricky, because `_validate_threshold` is just a wrapper around `Metadata.verify_delegate`, which does not tell us how many signatures that counted towards the threshold were verified. I suggest to ticketize this. It might actually be an interesting feature in python-tuf.

repository_service_tuf_worker/repository.py
lukpueh: Sure, I can add a comment. Probably also about the AND requirement?

tests/unit/tuf_repository_service_worker/test_repository.py
lukpueh: > We had a discussion with @kairoaraujo and @KAUTH regarding table testing done in this way and we are all with mixed feelings. The biggest issues I see with them:

Yeah, I figured. That's why I used the copy/paste approach for the first two main test cases, and only refactored on top of that. (Note though that the table does have a few extra test cases).
> 
> 1. it's harder to debug which test case fails when you see only `test_sign_metadata__update` with a specific assert that not always point you to the correct use case.

That's not necessarily true, if you look at the [log](https://github.com/repository-service-tuf/repository-service-tuf-worker/actions/runs/5890277848/job/15975064496?pr=355#step:6:195) you'll see that each call of the method is listed as separat test.

> 2. it could be a little complicated for new users.

🤷 I guess, that's debatable. If we expect devs to get familiar with pytest, monkey patching, pretend, mock, etc., we might as well ask them to read pytest docs for `parametrize`.  I personally find thousands of lines of testing code more intimidating. But yeah I agree with the intention of making testing code as little engineered as possible.

repository_service_tuf_worker/repository.py
MVrachev: Yes, that will be even better for people not familiar with TUF processes so much.

repository_service_tuf_worker/repository.py
MVrachev: I added a new issue for that: https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/367

MVrachev:
> Thanks for the review, @MVrachev! I'll push a commit for some of your inline suggestions. Regarding the things below:
> 
> > I saw that `ROLE_SIGNING` is used in many places where we have documented `SIGNING_ROLE`: https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#rstuf-repository-settings-configuration Do we want to update the docs or update our code? @lukpueh @kairoaraujo
> 
> Let's fix this in the docs, as agreed on slack.

Fixed by https://github.com/repository-service-tuf/repository-service-tuf/pull/433.

> > Also, what should we do with the `payload['role']` argument? If we decide to keep it we should make the code less opinionated about the role and not mention `root` everywhere (with the exception of the bootstrap case of course). I think there is a possibility we would introduce offline key(s) support for the `Targets` role in the future considering we want to give more flexibility on how users want to delegate trust, so I am in favor of keeping this argument for now.
> 
> See [#355 (comment)](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/355#discussion_r1304222629) for my feelings about `payload['role']`. If you folks insist on leaving it there, I can add a check for it.
> 
> Regarding making the function more generic, as you suggest inline, e.g. using `payload['role']` instead of a hardcoded "root", I am rather opposed. The function is simple enough to update, when we support other metadata types. Until then I don't see a point in making it more complicated than necessary or look more generic than it is.

I don't insist on making this change now, but given we already accept `role` argument in the API payload:
https://github.com/repository-service-tuf/repository-service-tuf-api/blob/407ffffef7f2c94cddcee2db6fd58cb256e3fb5e/repository_service_tuf_api/metadata.py#L152 
I think it's better we validate it at least and return a message if a non-root value is given.
Of course, another solution is changing the API itself.
Finally, of course, it's a solution to not do anything, but it's not ideal as we accept an argument we don't actually use.

The final question is about the testing. I think that this is a question that's debatable and we have different opinions.
I think it's better to not hurry to merge them directly, but try and come up with a decision with the rest of the team: @kairoaraujo and @KAUTH.
Given the circumstances and that we don't know how much time you will be around these weeks, I am open to merging this without the tests themselves, but we need to have an issue about this.

lukpueh:
I rebased and force pushed this branch to resolve some conflicts, and also added another commit to re-use parts of the "metadata update" task (see last commit).

For this feature to be useful / usable, we still need to change the "metadata update" task, to accept partially signed metadata. I have another commit for that which comes at ~ `65 additions and 44 deletions` (w/o tests).

Should I push it here, or open a separate PR?

kairoaraujo:
@lukpueh 
I'm ok with pushing it here. I think we made good progress, and it is partially reviewed.

lukpueh:
@kairoaraujo, @MVrachev: This is ready for another round of review. The most recent commit patches `metadata_update` task interface to allow partially signed root metadata, which can then be finalised via `sign_metadata`.

The big caveat is that `metadata_update` enters the signing event in any case if the the threshold is not reached. Also if there is not initial signature at all, or if there are invalid signatures, etc. Changing this would require a more invasive patch, which I think we should hold off until #367.

repository_service_tuf_worker/repository.py
MVrachev: It's more important to merge this soon than to generalize now.
When we need to allow this for other roles (such as `targets`) we will fix this then.

repository_service_tuf_worker/repository.py
MVrachev: It's more important to merge this soon than to generalize now.
When we need to allow this for other roles (such as targets) we will fix this then.

repository_service_tuf_worker/repository.py
MVrachev: It's more important to merge this soon than to generalize now.
When we need to allow this for other roles (such as targets) we will fix this then.

repository_service_tuf_worker/repository.py
```diff
@@ -1327,100 +1345,95 @@ def sign_metadata(
...
-                    "bootstrap": f"Root v{root_version} is pending signatures",
-                },
+        else:
+            # We need the "trusted root" when updating to a new root:
+            # - signature could come from a key, which is only in the trusted
+            #   root, OR from a key, which is only in the new root
+            # - threshold must validate with the threshold of keys as defined
+            #   in the trusted root AND as defined in the new root
+            trusted_root = self._storage_backend.get("root")
+            trusted_signature = self._validate_signature(
```
MVrachev, line 1417: Nit for better readability:
```suggestion
            is_trusted_signature = self._validate_signature(
```

repository_service_tuf_worker/repository.py
```diff
@@ -1327,100 +1345,95 @@ def sign_metadata(
...
-        self._bootstrap_finalize(root, bootstrap_task_id)
-        return self._task_result(
-            TaskName.SIGN_METADATA,
-            True,
-            {
-                "message": "Signature Processed",
-                "bootstrap": "Bootstrap Finished",
-            },
-        )
+            new_signature = self._validate_signature(root, signature)
```
MVrachev, line 1420: Nit for better readability:
```suggestion
            is_new_signature = self._validate_signature(root, signature)
```

kairoaraujo:
> I will review the code today, but for testing, it's good if we have [repository-service-tuf/repository-service-tuf#515](https://github.com/repository-service-tuf/repository-service-tuf/pull/515).
> Do we want to work on merging this without functional tests first or do we want to wait for them @kairoaraujo, @lukpueh?

I'm running the tests to check if it doesn't affect the fully signed metadata update.
I don't see issues so far, so merging this has no problems.

tests/unit/tuf_repository_service_worker/test_repository.py
MVrachev: We need to decide about this test and methodology @kairoaraujo and @lukpueh.
Do we want to meet or discuss it here?

PS: Now when I read it it's not immediately clear to me which test case is testing what.
Maybe if there was some kind of a title of a test case would be better.
Something familiar to what we did in python-tuf here: https://github.com/theupdateframework/python-tuf/blob/038ecd65dc6b74e9d0306d3c57c3e7d621155e47/tests/test_updater_key_rotations.py#L92

MVrachev (commented):
The main code is clean, elegant and easy to read.
Left a couple of small nits about it.

We still need to discuss:
- do we want to merge this without FT tests?
- what to do with the `parametrize` test?

lukpueh:
> We need to decide about this test and methodology @kairoaraujo and @lukpueh.

Let me know what it should be. If you are fine with my approach, I can add titles as Martin suggests. Otherwise, I can copy paste. Note that the difference is ~140 vs. ~450 LOC

lukpueh:
> In a meeting with @kairoaraujo and our new contributors @enyinna1234 and @ivanayov they read the parametrize test and shared it was difficult for them to follow the test.

Okay, that's a good argument. One final note, though ... and then I'll gladly change it. Is it really the usage of `parametrize` that makes the test difficult to follow? 😄 

As I said before, I think that nested functions, "fake" variables, `pretend`, `stub`, `call_recorder`, `monkeypatch`, `lambda`, `iter`, etc. set the bar for complicated constructs fairly high. It took me at least quite a while to understand their usage in other copy-paste style tests, when writing this one.

Would code documentation like this clarify the use of `parametrize` (it's really just an invisible for loop) ...?
```
# `parametrize` calls the test function for each "row" in the test data table. 
# The fields in a row are passed as arguments to the test function. 
#
#   payload_patch: input part for tested function (`sign_metadata`) 
#   validation_results: mock return value of called helper functions
#   details: expected return value part of test function
#   status: expected return value part of test function
```

I am very much open for renames and restructures of these variables, e.g. `validation_results`  could easily be split in two variables and/or `details` and `status` could be combined in one. There could also be a title field, which describes each test scenario, as Martin suggested.

lukpueh:
Force pushed w/o change to rebase on `main` and added two new commits:
- var renames, as requested by @MVrachev 
- copy tests instead of using `@parametrize`, as requested by all :)

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -3428,31 +3412,312 @@ def fake_get_fresh(key):
...
+        test_repo._root_metadata_update = pretend.call_recorder(
+            lambda a: {
+                "message": "Metadata Update Processed",
+                "role": "root",
+            }
+        )
+
+        # Use `next` below to mock subsequent calls
+        fake_signature_result = iter((False, True))
+        fake_threshold_result = iter((False, True))
```
MVrachev, line 3554: Interesting way of mocking return type.
I like it better than compared to using `MagicMock` as we did until now when we have to mock a different behavior of two calls.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -3310,83 +3310,67 @@ def fake_get_fresh(key):
...
+        test_repo.write_repository_settings = pretend.call_recorder(
+            lambda *a: "fake"
+        )
+
+        test_repo._root_metadata_update = pretend.call_recorder(
+            lambda a: {
+                "message": "Metadata Update Processed",
+                "role": "root",
+            }
+        )
```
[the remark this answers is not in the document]
lukpueh: Excellent observation, @MVrachev! I removed (hopefully) all unnecessary mocks from my tests.

MVrachev (approved):
Reapprove to run functional tests with pr https://github.com/repository-service-tuf/repository-service-tuf/pull/515

kairoaraujo (approved):

PR #354, opened 2023-08-13
fix: Automatic bump online roles, meta update

kairoaraujo:
It fixes the `_update_snapshot` to update the Snapshot Meta with the latest versions.

It was missing in the loop to add the latest role json version. It also improves the UT to check the changes in the snapshot version and the meta updates to avoid regression.

Closes #353

PR #357, opened 2023-08-14
Add in UT a check if timestamp version is updated

kairoaraujo:
This commit is a follow-up to the PR #354 

It adds a test to check if the `timestamp` version is updated to avoid regression on changes in RSTUF.

### DIFF
--- a/tests/unit/tuf_repository_service_worker/test_repository.py
+++ b/tests/unit/tuf_repository_service_worker/test_repository.py
@@ -310,8 +310,8 @@ class TestMetadataRepository:
             lambda *a: mocked_timestamp
         )
 
-        def fake__bump_and_persist(*a, **kw):
-            mocked_timestamp.signed.version += 1
+        def fake__bump_and_persist(md, role, **kw):
+            md.signed.version += 1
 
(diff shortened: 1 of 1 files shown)


tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -295,22 +295,31 @@ def test__bump_and_persist_without_persist(self, test_repo):
...
+            signed=pretend.stub(snapshot_meta=2, version=timestamp_version)
+        )
         test_repo._storage_backend.get = pretend.call_recorder(
             lambda *a: mocked_timestamp
         )
-        test_repo._bump_and_persist = pretend.call_recorder(lambda *a: None)
 
-        result = test_repo._update_timestamp(snapshot_version)
+        def fake__bump_and_persist(*a, **kw):
+            mocked_timestamp.signed.version += 1
```
MVrachev, line 314: It's not a great idea or a good practice to access a variable outside the function when the same variable is given as an argument.
Better do it as you did in https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/354 where you bump the metadata passed as an argument.

MVrachev (approved):

PR #390, opened 2023-09-15
Stop calling refresh_settings on each task

MVrachev:
Currently, we are calling refresh_settings many times:
1. a couple of times when we set up SQL, Storage Backend, KeyVault and timeout when initializing MetadataRepository
2. once per task we call refresh_settings

In a discussion with Kairo, we reevaluated this design and realized that if there is a change that needs to be made then the user needs to update the environmental variables which will require a restart of all worker containers and as a result, refresh_settings will be called for sure. This means that us calling refresh_settings on each task or during MetadataRepository initialization is redundant as nothing can be changed from the settings without the requirement to restart the containers, otherwise, an error will be raised.

### DIFF
--- a/app.py
+++ b/app.py
@@ -75,7 +75,6 @@ def repository_service_tuf_worker(
     self,
     action: str,
     payload: Optional[Dict[str, Any]] = None,
-    refresh_settings: Optional[bool] = True,
 ):
     """
     Repository Service for TUF Metadata Worker main Celery consumer.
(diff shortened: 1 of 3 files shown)


kairoaraujo (approved):

kairoaraujo (approved):

kairoaraujo (approved):

kairoaraujo (approved):

kairoaraujo (approved):

kairoaraujo (dismissed):

kairoaraujo (approved):

kairoaraujo (approved):

kairoaraujo (approved):

MVrachev:
Superceeded by #395.
I added a new pr, so we can run the functional tests on approval as we are yet to solve this problem.

PR #397, opened 2023-09-20
Support multiple envs for setting

kairoaraujo:
- The old `repository.refresh_settings` handling the IServices
  configuration was moved to the `interfaces`
- Now the way to importa a `I{Services}` uses directly the dynaconf
- Adds better error mensages for wrong configuration
- Support a service has one setting with multiple environment variables
  what is helpful for third-part libraries that uses environment vars


MVrachev:
Superceeds: https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/389.

PR #360, opened 2023-08-15
AWS S3 Storage Backend Service implementation

kairoaraujo:
### Prepare RSTUF development environment

  - Add boto3 to use AWS services
  - Add awswrangler to parse objects in the bucket
  - Add localstack to enable a development in the docker-compose.yml
  - Add localstack-init in Makefile to enable developers to initiate the
    services configuration
  
### Implements the AWS S3 Storage Backend Service

It uses the official `boto3` library for AWS Sessions, Client and
Resources and also the `awswrangler` to query the AWS S3 metadata.

It implements all the unit tests covering 100% coverage.

Updates the RSTUF Worker documentation adding a better format for
Storage Backend and KeyVault Backend.

PR #407, opened 2023-10-03
Use I(KeyVault|Storage).cofigure as factory method 

### LINKED ISSUE #401: Task: transform "IStorage" and "IKeyvault" configure method to a constructor
### What is the task about?

I suggest we change configure from `classmethod` to a regular `self` method and instantiate a new object of type `IStorage`. 

I propose the same change for `IKeyVault`.

### Reasons to do that

1. We already call `configure` just before actually instantiating an object:
- `IStorage`: https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/0e64818712e042a4d8353da0ddb402d2dbaae7a1/repository_service_tuf_worker/interfaces.py#L163
- `IKeyVault`: https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/0e64818712e042a4d8353da0ddb402d2dbaae7a1/repository_service_tuf_worker/interfaces.py#L168

2. We have to allocate resources and connections twice: once at `configure` and a second time during an actual initialization leading to:
- duplicating code in both `configure` and `__init__`.
- for online storage and keyvault options it leads to double the connections and requests

3. The name itself `configure` implies you are configuring something, but in the end it doesn't result to an object.
This means that we can keep the name `configure` and make it a constructor.
If we do that this will a similar model to what `python-tuf` metadata API uses with their [from_dict](https://github.com/theupdateframework/python-tuf/blob/b7c956cd01213ddc9a2e2605a643536608e85b13/tuf/api/metadata.py#L164) and `securesystemslib` signer API method [from_priv_key_uri](https://github.com/secure-systems-lab/securesystemslib/blob/d7c23cce61b650685dff3533af04d11c9fa48c74/securesystemslib/signer/_signer.py#L90).

### Code of Conduct

MVrachev:
Fixes #401 

Transform the I(KeyVault|Storage).cofigure as factory method meaning
it will be used as a constructor.
This will allow removing duplicating code from `__init__` or `configure`.

Additionally, I used the opportunity to remove the "kwargs" variable
and operations inside interfaces._setup_service_dynaconf() as we
don't use it for anything. 

As a continuation, I was able also to remove
ServerSettings.argument field from ServerSettings as again it was not
used anywhere any more.

Finally, I used `configure` as a constructor to remove the calls
of `_raw_key_parser` inside `LocalKeyVault.get()` as the result
of those calls will be the same during the container's lifetime.

### DIFF
--- a/repository_service_tuf_worker/interfaces.py
+++ b/repository_service_tuf_worker/interfaces.py
@@ -26,7 +26,7 @@ class IKeyVault(ABC):
     @abstractmethod
     def configure(cls, settings: Dynaconf) -> "IKeyVault":
         """
-        Run actions to check and configure the service using the settings.
+        Run actions to test, configure and create object using the settings.
         """
         pass  # pragma: no cover
 
@@ -56,7 +56,7 @@ class IStorage(ABC):
     @abstractmethod
     def configure(cls, settings: Dynaconf) -> "IStorage":
         """
-        Run actions to test, configure using the settings.
+        Run actions to test, configure and create object using the settings.
         """
         raise NotImplementedError  # pragma: no cover
 
--- a/repository_service_tuf_worker/interfaces.py
+++ b/repository_service_tuf_worker/interfaces.py
@@ -17,7 +17,6 @@ class ServiceSettings:
     """Dataclass for service settings."""
 
     names: List[str]
-    argument: str
     required: bool
     default: Optional[Any] = None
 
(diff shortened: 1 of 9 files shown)


MVrachev:
Let's merge this one after #360.
After that I can update this pr to support `AWS S3`.

kairoaraujo (commented):
I like the improvement added to the IServices.
I have two small comments.

kairoaraujo (approved):

PR #424, opened 2023-11-22
Issue #352 - Replacing mock datetime by fake_datetime fixture

### LINKED ISSUE #352: Task: Replace mock `datetime` by `fake_datetime` fixture
### What is the task about?

We recently implemented a new fixture that automatically mocks the datetime. 

This task is about replacing the monkeypatch mock by the fixture

Remove mocks like this

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/2e071289e5353d5f47001e35942f477f82f4074f/tests/unit/tuf_repository_service_worker/test_repository.py#L233-L239

By a the fixture `mocked_datetime` which does all 5 lines 😀

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/2e071289e5353d5f47001e35942f477f82f4074f/tests/unit/tuf_repository_service_worker/test_repository.py#L770-L771

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

fsavoia:
# Description

Fixes #352 

# Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

# Additional requirements
- [x] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

# Code of Conduct
By submitting this PR, you agree to follow our [Code of Conduct](CODE_OF_CONDUCT.rst).


tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -230,23 +230,17 @@ def test_bump_expiry(self, monkeypatch, test_repo):
...
+            signed=pretend.stub(expires=fake_datetime.now()),
         )
 
         result = test_repo._bump_expiry(fake_role, "root")
         assert result is None
         assert fake_role.signed.expires == datetime.datetime(
             2023, 6, 15, 9, 5, 1
         )
-        assert fake_datetime.now.calls == [pretend.call()]
+        assert fake_datetime.now.calls == [pretend.call(), pretend.call()]
```
MVrachev, line 243: I see that by introducing a new `mocked_date` adds a new `pretend.call()` because of line 235.
I don't see how this can be fixed.
Do you @kairoaraujo?

MVrachev (commented):
When you run `make docs` locally we are also updating the png files:
- `docs/source/_static/repository-service-tuf-worker-C1.png`
- `docs/source/_static/repository-service-tuf-worker-C2.png`
- `docs/source/_static/repository-service-tuf-worker-C3.png`

I was and I am still puzzled do we want to do that?
Before in discussions @kairoaraujo said that it's better to remove them from a commit, but why? Maybe it's because `plantuml` (used to generate those pictures) version is not controlled by us and can generate different binary files depending on the version used.
What would you say @kairoaraujo?

kairoaraujo:
> I was and I am still puzzled do we want to do that? Before in discussions @kairoaraujo said that it's better to remove them from a commit, but why? Maybe it's because `plantuml` (used to generate those pictures) version is not controlled by us and can generate different binary files depending on the version used. What would you say @kairoaraujo?

Yes, IMO It is _nit_ remove it from the commit as the reviewers are aware about this issue of rebuilding the plantuml images. We should create a issue to solve it.

tests/unit/tuf_repository_service_worker/test_repository.py
kairoaraujo: I would say that we could remove the `fake_datetime.now.calls` check.
I think it is less confusing for contributors.

fsavoia: Thank you for your answer. So I deleted this assert.

tests/unit/tuf_repository_service_worker/test_repository.py
```diff
@@ -230,23 +230,16 @@ def test_bump_expiry(self, monkeypatch, test_repo):
...
             lambda *a, **kw: fake_settings,
         )
-        fake_time = datetime.datetime(2019, 6, 16, 9, 5, 1)
-        fake_datetime = pretend.stub(
-            now=pretend.call_recorder(lambda: fake_time)
-        )
-        monkeypatch.setattr(
-            "repository_service_tuf_worker.repository.datetime", fake_datetime
-        )
+        fake_datetime = mocked_datetime
```
MVrachev, line 233: I don't see why you renamed `fake_datetime` to `mocked_datetime` everywhere.
I can understand if it's a single instance where a line goes above 79 characters if you use `mocked_datetime` instead and you decide to rename it, but why do you do it everywhere?
I think it's clear that if something is mocked then it's faked.

fsavoia: I agree with you. Makes sense to use `mocked_datetime` directly. I just used `fake_datetime` because I got a reference that was already in the code, like in the function `test_sign_metadata_invalid_role_type` (current line 3145). So I'm going to change now, including this reference.

MVrachev (approved):
Nice work!
@kairoaraujo I am approving, do you want to remove the  changed binary `png` files or we are good to go?

MVrachev (approved):

kairoaraujo:
> Nice work! @kairoaraujo I am approving, do you want to remove the changed binary `png` files or we are good to go?

It is ok replace it.

Issue #419, opened 2023-11-13
Use securesystemslib signer API as intended


lukpueh:
### What do you want to share with us?

RSTUF does not use the securesystemslib Signer API as intended. Understandably, because the preferred usage pattern is not self-explanatory and not yet well documented.

But the API is actually quite simple and powerful. The underlying assumptions are that signers are authorized prior to being used, and that the authorization is known to the application when loading the signer. Both is true for RSTUF.

To load any signer for a public key, we can then use the generic `Signer.from_priv_key_uri` interface and pass it a URI to the private key material. The URI can e.g. contain a path to a local key file, or a keyid for a cloud kms, etc.

The nice thing about this usage pattern is that the application code does not need to know about the specific signer implementation.

Here is an implementation for a generic SignerStore to load, cache and use arbitrary signers for online signing in the RSTUF worker: https://github.com/repository-service-tuf/repository-service-tuf-worker/commit/f3d1d6323ea6f3c7d460e0f5efcb5faae9cbc58b

Implementation Notes:

- Includes signer-specific code in order to parse the existing signer-specific config into URIs. If we store URIs directly in config, we can make the implementation completely signer agnostic and further simplify the implementation.

- Works around the missing "public key" to "signer" mapping in config, by just picking the first signer in the store for any public key. This is similar to what `ILocalKeyVault` currently does, and should be fixed.

If we choose the proposed approach, I recommend updating the signer config format to list <keyid>=<signer uri> pairs and to automate the config creation in the RSTUF CLI.

An alternative approach, which works without additional config at all, is to include signer URIs with the public keys in root metadata, as custom fields. This is what tuf-on-ci does to configure online signers.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #427, opened 2023-12-04
Add simple online signer store (wip)

lukpueh:
- Can be used to load/cache/use any signer from securesystemslib Signer API, including custom signers.
- Should replace IKeyVault and related classes.
- Details in #419

### DIFF
--- a/repository_service_tuf_worker/_signer.py
+++ b/repository_service_tuf_worker/_signer.py
@@ -1,86 +1,33 @@
-from base64 import b64decode
-from hashlib import blake2b
-from pathlib import Path
-
-from securesystemslib.signer import Key, SecretsHandler, Signer
-
-from repository_service_tuf_worker import Dynaconf
+from securesystemslib.signer import Key, Signer
 
 
 class SignerStore:
     """Generic signer store.
 
-    - Loads per-signer info (uri, secrets handler) from config on `init`.
-    - Loads signer for pub key on `get`.
+    Provides method to load and cache signer for passed public key, using a URI
+    configured in a custom field of the passed key.
 
-    FIXME: include keyid in signer config to map config to public keys
     """
 
-    def __init__(self, settings: Dynaconf):
+    def __init__(self):
         self._signers: dict[str, Signer] = {}
-        self._signer_infos: dict[str, tuple[str, SecretsHandler]]
-
-        if settings.KEYVAULT_BACKEND == "LocalKeyVault":
-            for signer_info in settings.LOCAL_KEYVAULT_KEYS.split(":"):
-                keyid, uri, secrets_handler = self._parse_local(
-                    signer_info, settings.LOCAL_KEYVAULT_PATH
-                )
-
-            self._signer_infos[keyid] = (uri, secrets_handler)
-
-    @staticmethod
-    def _parse_local(
-        signer_info: str, signer_dir: str
-    ) -> tuple[str, str, SecretsHandler]:
-        """Parses keyid, uri and secrets handler from single signer info chunk.
-
-        If chunk contains key data, it is written to a file.
-        Colon-separated chunks can be found in RSTUF_LOCAL_KEYVAULT_KEYS.
-
-        FIXME: encoding is overly complex, each chunk can be one of:
-            1. path to key file , password, type (optional)
-            2. base64 encoded key data, password, type (optional)
-            3. path to a secret, which contains (1)
-
-        I recommend to pick one good way to configure file-based keys only
-        """
-        # Case-handle signer_info is in a secret
-        if signer_info.startswith("/run/secrets/"):
-            with open(signer_info) as f:
-                signer_info = f.read().rstrip("\n")
-
-        # Unpack signer_info, ignore irrelevant optional 'keytype' (*_)
-        # FIXME: 'keytype' in public key is authoritative, no need to config
-        filename, password, *_ = signer_info.split(",")
-
-        # Case-handle signer_info contains key data
-        if filename.startswith("base64|"):
-            key_data = b64decode(filename[len("base64|") :])
-            filename = blake2b(key_data, digest_size=16)
-            with open(Path(signer_dir) / filename, "wb") as f:
-                f.write(key_data)
-
-        # FIXME: we can't use this keyid because it does not map to a pubkey.
-        keyid = filename
-
-        path = f"{signer_dir}/{filename}"
-        uri = f"file:{path}?encrypted=true"
-
-        # FIXME: what is the benefit of using a 'password', which is passed
-        # along with the secret key data?
-        secrets_handler = lambda name: password
-
-        return keyid, uri, secrets_handler
 
     def get(self, key: Key) -> Signer:
-        """Return signer for passed key."""
+        """Return signer for passed key.
+
+        NOTE: the passed `key` is expected to have an "x-rstuf-online-key-uri"
+        field in its `unrecognized_fields` attribute. This uri is used with the
+        generic `Signer.from_priv_key_uri` interface, and, along with the
+        public key, it encodes all information necessary to load the signer.
+
+        Additional required information, e.g. to authenticate with a Cloud KMS,
+        may be provided via signer type specific environment variables.
+
+        """
 
         # If signer not in cache, load it using config
         if key.keyid not in self._signers:
-            # FIXME: provide keyid mapping and use it here
-            # uri, sec = self._signer_infos[key.keyid]
-            uri, sec = self._signer_infos.values()[0]
-
-            self._signers[key.keyid] = Signer.from_priv_key_uri(uri, key, sec)
+            uri = key.unrecognized_fields["x-rstuf-online-key-uri"]
+            self._signers[key.keyid] = Signer.from_priv_key_uri(uri, key)
 
         return self._signers[key.keyid]
--- a/repository_service_tuf_worker/_signer.py
+++ b/repository_service_tuf_worker/_signer.py
@@ -1,4 +1,49 @@
-from securesystemslib.signer import Key, Signer
+import os
+from pathlib import Path
+from typing import Optional
+
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
+from securesystemslib.signer import (
+    SIGNER_FOR_URI_SCHEME,
+    CryptoSigner,
+    Key,
+    SecretsHandler,
+    Signer,
+)
+
+
+class FileSigner(CryptoSigner):
+    """File-based signer implementation.
+
+    Overrides `CryptoSigner.from_priv_key_uri` to load private key file from a
+
+    * directory, defined in envvar: `RSTUF_ONLINE_KEY_DIR`, and a
+    * file name, defined in the passed uri: `fn:<file name>`
+
+    """
+
+    SCHEME = "fn"
+    DIR_VAR = "RSTUF_ONLINE_KEY_DIR"
+
+    @classmethod
+    def from_priv_key_uri(
+        cls,
+        priv_key_uri: str,
+        public_key: Key,
+        secrets_handler: Optional[SecretsHandler] = None,
+    ) -> "FileSigner":
+        _, _, file_name = priv_key_uri.partition(":")
+        dir_ = os.environ[cls.DIR_VAR]
+
+        with open(Path(dir_, file_name), "rb") as f:
+            private_pem = f.read()
+
+        private_key = load_pem_private_key(private_pem, None)
+        return FileSigner(private_key, public_key)
+
+
+# Register signer for scheme for usage via generic `Signer.from_priv_key_uri`
+SIGNER_FOR_URI_SCHEME.update({FileSigner.SCHEME: FileSigner})
 
 
 class SignerStore:
--- a/repository_service_tuf_worker/_signer.py
+++ b/repository_service_tuf_worker/_signer.py
@@ -42,8 +42,34 @@ class FileSigner(CryptoSigner):
         return FileSigner(private_key, public_key)
 
 
+class EnvSigner(CryptoSigner):
+    """Environment variable -based signer implementation.
+
+    Overrides `CryptoSigner.from_priv_key_uri` to load private key from an
+    environment variable defined in the passed uri:
+
+    `env:<environment variable name>`
+
+    """
+
+    SCHEME = "env"
+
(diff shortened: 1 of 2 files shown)


MVrachev (commented):
I will add some questions and I will enumerate them, so you can address them:
1.  Can you describe why this is a better implementation of what we currently do - one child class per signer?
2. The next important question is what will it take to use this new approach to add support for a new signer as in my case for AWSSigner.
What I see is that: I will need to add a section in `__init__`  specificy for AWSSigner where I will need to import my AWSKey, import the signer, use it and make sure it's working
3. How do we make sure that the private keys we have loaded through `from_priv_key_uri` can work with the public key that has been used to load that private key?
I ask as I don't see such a code in securesystemslib.

lukpueh:
> I will add some questions and I will enumerate them, so you can address them:

Great questions!


> 1. Can you describe why this is a better implementation of what we currently do - one child class per signer?

It is easier, 
The , we have a redundant layer of abstraction over individual Signer implementations 



> 2. The next important question is what will it take to use this new approach to add support for a new signer as in my case for AWSSigner.
>    What I see is that: I will need to add a section in `__init__`  specificy for AWSSigner where I will need to import my AWSKey, import the signer, use it and make sure it's working

The current work we do in `__init__` (and `_parse_local`) is just a bridge to convert the current config t private key URIs understood by `Signer`.

If we directly configure `<keyid>=<signer uri>` pairs, `SignerStore.get` will just work for any signer implementation known to the securesystemslib Signer API (modulo different secrets handler, which we may not need at all for online signing).



> 3. How do we make sure that the private keys we have loaded through `from_priv_key_uri` can work with the public key that has been used to load that private key?
>    I ask as I don't see such a code in securesystemslib.

The most reliable way to see if they belong together is to create a signature and verify it, which we'll do anyway when bootstrapping/updating metadata. Otherwise, it should be okay to just rely on the mapping by keyid.

lukpueh:
> I will add some questions and I will enumerate them, so you can address them:

Great questions!

> 1. Can you describe why this is a better implementation of what we currently do - one child class per signer?

We currently have a redundant abstraction layer over individual signer implementations. Why implement a `LocalKeyVault` class, if securesystemslib already gives you a `CryptoSigner` class?

Moreover, the `IKeyVault` interface, is not as powerful as `Signer`. The latter gives you a method, to transparently load any signer based on a configured URI. This means you don't have to deal with specific signer implementations in code at all.

Maybe the advantage becomes more apparent below:

> 2. The next important question is what will it take to use this new approach to add support for a new signer as in my case for AWSSigner.
>    What I see is that: I will need to add a section in `__init__`  specificy for AWSSigner where I will need to import my AWSKey, import the signer, use it and make sure it's working

The work we currently do in `__init__` (and `_parse_local`) is just a bridge to convert the existing config to a URI, which is  compatible with the Signer API.

If we'd directly configure `<keyid>=<signer uri>` pairs, `SignerStore.get` would just work for any signer implementation known to the securesystemslib Signer API, without prior case handling.

For AWS this means:
1. user writes  `<local keyid>=awskms:<aws keyid>` and [aws envvars](https://github.com/secure-systems-lab/securesystemslib/pull/627#issuecomment-1694216676) to online key signer config
2. worker exposes aws envvars in signer environment: this could be done in `SignerStore.__init__` but also independently
3. SignerStore caches configured `<local keyid>=awskms:<aws keyid>` on init: if all signers are configured that way, parsing becomes generic
4. caveat: in my current implementation, we also create signer class specific secrets handlers. IMO online signing shouldn't need a secrets handler at all (but we can discuss this separately)
5. Finally, `SignerStore.get` transparently returns a specific signer for the passed the key (keyid).


> 3. How do we make sure that the private keys we have loaded through `from_priv_key_uri` can work with the public key that has been used to load that private key?
>    I ask as I don't see such a code in securesystemslib.

The most reliable way to see if they belong together is to create a signature and verify it, which we'll do anyway when bootstrapping/updating metadata. Otherwise, it should be okay to just rely on the mapping by keyid.

kairoaraujo:
I understand that the significant change here is that we no longer need to create the IKeyVault Service.

The background here, @MVrachev is that we started with securesystemslib (sslib) in [RSTUF when it was on version 0.23.0](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/v0.0.1a1/requirements.txt).

At this moment, sslib supported only files for signing, so the idea was to use it for the IKeyVault service `LocalKeyVault` only, and I assumed that we should create our signers for HSM, Cloud Services, etc.

The sslib has greatly improved and now supports all these signers; we can simplify it and relay it on securesystemslib.
For any future implementation, as HashiCorp Vault, we should contribute with sslib.

In the RSTUF, we need to keep the user experience of configuring the containers easily, implementing the environment variables (`RSTUF_*`), supporting secrets, custom configurations, etc.

lukpueh:
> In the RSTUF, we need to keep the user experience of configuring the containers easily, implementing the environment variables (RSTUF_*), supporting secrets, custom configurations, etc.

Is this a response to something I said? Regardless, I have some comments... :)

**Re: container config UX...**

Agreed, this should be super easy. Asking the user to come up with config like `<long hex string>=<custom uri format>` is not great. I suggested to automate this. E.g. when the user adds the public key in root metadata via the RSTUF admin cli, we can ask them to provide the necessary information for the tool to construct the URI, e.g. *"Enter the path to the online private key in the container:"*. Or, the other way around, we tell them where to write the key to, e.g. *"Make sure the key is available at /run/secrets/KEYID in the container!"*.

Then the CLI could print the constructed `<keyid>=<uri>` config entry and instruct the user to include it in the worker config. But this is still not great. The user already has to configure the key itself, configuring the location too, does not seem user friendly.

Alternatively, and this is what tuf-on-ci does, we could just include the uri with the public key in tuf metadata. Then the user does not have to configure the key information at all, just the key itself.

I think we should go for that.
 
**Re: environment variables (`RSTUF_*`)**
We discussed this with Jussi, but I can't remember why it is important that all environment variables have an RSTUF_ prefix? If we are dealing with a 3rd party env var, which RSTUF just "relays", why not use the original name? E.g. what's the benefit of asking the user to write `AWS_ACCESS_KEY_ID` as `RSTUF_AWS_ACCESS_KEY_ID`, just so that RSTUF internally has to then strip the prefix, before rewriting it to the environment?

**Re: supporting secrets, custom configurations, etc.**
Yeah, about that, I wonder if we really need to support multiple different ways of configuring local online keys. Especially, in the light of other signers in securesystemslib, which seem better suited for online signing (gcp, aws, azure). 

Currently, you can configure a private key via `RSTUF_LOCAL_KEYVAULT_KEYS` as one of:
- path to file with key data
- path to file with path to file with key data
- key data

I suggest to allow either path to file with key data or key data itself.

kairoaraujo:
> > In the RSTUF, we need to keep the user experience of configuring the containers easily, implementing the environment variables (RSTUF_*), supporting secrets, custom configurations, etc.
> 
> Is this a response to something I said? Regardless, I have some comments... :)
> 
> **Re: container config UX...**
> 
> Agreed, this should be super easy. Asking the user to come up with config like `<long hex string>=<custom uri format>` is not great. I suggested to automate this. E.g. when the user adds the public key in root metadata via the RSTUF admin cli, we can ask them to provide the necessary information for the tool to construct the URI, e.g. _"Enter the path to the online private key in the container:"_. Or, the other way around, we tell them where to write the key to, e.g. _"Make sure the key is available at /run/secrets/KEYID in the container!"_.

Not clear here, but I believe you're mentioning the use case for type LocalKeyVault, which we agree shouldn't be most used in production deployments (but will have anyway).
We need to make it easy and I agree we can simplify if there are ways.
The case for supporting multiple cases here. While deploying the container, there are a lot of chicken and egg problems.
Specifically, LocalKeyStorage is the first-experience-user trying RSTUF.

> Then the CLI could print the constructed `<keyid>=<uri>` config entry and instruct the user to include it in the worker config. But this is still not great. The user already has to configure the key itself, configuring the location too, does not seem user friendly.
> Alternatively, and this is what tuf-on-ci does, we could just include the uri with the public key in tuf metadata. Then the user does not have to configure the key information at all, just the key itself.
> 
> I think we should go for that.

Should we make the user obligated to use the CLI? Now, it is more of an interface and interface example for RSTUF API.
But it is a decision we should make.
I like the idea that an organization can create its own interface for RSTUF API, including its own process.

Remember that the RSTUF deployment can have a different setup as it uses containers. There is a difference from tuf-on-ci here.
The users must deal with deployment, which means configuring containers using compose and manifests depending on the strategy and environment.
> 
> **Re: environment variables (`RSTUF_*`)** We discussed this with Jussi, but I can't remember why it is important that all environment variables have an RSTUF_ prefix? If we are dealing with a 3rd party env var, which RSTUF just "relays", why not use the original name? E.g. what's the benefit of asking the user to write `AWS_ACCESS_KEY_ID` as `RSTUF_AWS_ACCESS_KEY_ID`, just so that RSTUF internally has to then strip the prefix, before rewriting it to the environment?

Here, I disagree.
First, because we are saying to the user to go to `securesystemslib`, that will say got to `boto3` and use their environment variable. It is a bit distributed experience. Imagine doing it for SQLAlchemy, and Redis,  we use all those libraries to build the service.

Secondly, it is a standard. Here are some examples: 
(See the environment variables)
https://hub.docker.com/_/postgres
https://hub.docker.com/r/bitnami/kafka
https://hub.docker.com/_/mysql

> **Re: supporting secrets, custom configurations, etc.** Yeah, about that, I wonder if we really need to support multiple different ways of configuring local online keys. Especially, in the light of other signers in securesystemslib, which seem better suited for online signing (gcp, aws, azure).
> 
> Currently, you can configure a private key via `RSTUF_LOCAL_KEYVAULT_KEYS` as one of:
> 
> * path to file with key data
> * path to file with path to file with key data
> * key data
> I suggest to allow either path to file with key data or key data itself.

It sounds strange, but I think we support the two common use cases for containers with the constraint that it is a volume always.
 * path to the key: Which covers the file that is available or stored in a secret
 * base64 encoded key content: Which covers the file will be saved in the volume
    - It supports be given as secrets as well (base64)

lukpueh:
>  While deploying the container, there are a lot of chicken and egg problems. Specifically, LocalKeyStorage is the first-experience-user trying RSTUF.

What problems do you mean and how are they related to my proposal? 

> Should we make the user obligated to use the CLI?

No, and we don't need to. An organisation can very well build their own tooling to create TUF metadata, with private key uris included in public keys.

> The users must deal with deployment, which means configuring containers using compose and manifests depending on the strategy and environment.

Sure, that doesn't change with my proposal.  As before, the user still needs to get the private key somehow into the system. The difference is that now the user only needs to configure the key (e.g. via mounting a volume with the key file), and the information where to get the key from is included transparently via TUF metadata. Whereas previously the user had to configure both, the key and the information where to load it from.
 

> Here, I disagree. First, because we are saying to the user to go to `securesystemslib`, that will say got to `boto3` and use their environment variable. It is a bit distributed experience. Imagine doing it for SQLAlchemy, and Redis, we use all those libraries to build the service.

I would argue that it is already a distributed experience, because the user will likely manage their aws key via aws tooling.

It seems okay to say in user docs something like, *"if you are using aws, make sure to configure these aws env vars: [link to aws env var docs]"*.

Alternatively, you'd have to say something like, *"if you are using aws make sure to configure these rstuf env vars, which correspond to these aws env vars but w/o the RSTUF_ prefix: [link to aws env var docs]"*. :P 

But yeah, I won't insist. It's okay with me to just re-export them.

> > * path to file with key data
> > * path to file with path to file with key data
> > * key data
> >   I suggest to allow either path to file with key data or key data itself.
> 
> It sounds strange, but I think we support the two common use cases for containers with the constraint that it is a volume always.
> 
> * path to the key: Which covers the file that is available or stored in a secret
> * base64 encoded key content: Which covers the file will be saved in the volume
>   
>   * It supports be given as secrets as well (base64)

Do you disagree that we currently support the 3 ways of configuring local keys above? Also, my question was, if the user really needs all that flexibility.

lukpueh:
@kairoaraujo, @MVrachev, I updated the PR to implement what we discussed:

d079add rewrites the signer store to remove any custom key config parsing and instead uses a key URI, which can be attached to the public key. This solves the problem of mapping public keys to private keys at signing time, works for any registered Signer implementation, and makes the signing code extremely simple.

The other two commits (39ea16c, 68ae260) implement two custom signer loaders. One for relative path URIs, where a base bath is expected as environment variable (as you requested). And one for URIs, which point to environment variable names, from which the key can be loaded.

As you can see, they only need to be registered in the Signer API's dispatch table for `SignerStore.get`  to use them. No other change is needed. 

Btw. I mostly removed the config parsing code for emphasis (to show how the code becomes even simpler). If we want to stay backwards compatible we can easily revert that and make the SignerStore use the old-style config if it exists, but expect URIs on keys otherwise.

MVrachev:
> @kairoaraujo, @MVrachev, I updated the PR to implement what we discussed:
> 
> [d079add](https://github.com/repository-service-tuf/repository-service-tuf-worker/commit/d079add333916b9d69b55dc2a837d52a9b62a42c) rewrites the signer store to remove any custom key config parsing and instead uses a key URI, which can be attached to the public key. This solves the problem of mapping public keys to private keys at signing time, works for any registered Signer implementation, and makes the signing code extremely simple.
> 
> The other two commits ([39ea16c](https://github.com/repository-service-tuf/repository-service-tuf-worker/commit/39ea16cb18a3dcdfd14a6cc324e3c4d3188a1ee7), [68ae260](https://github.com/repository-service-tuf/repository-service-tuf-worker/commit/68ae260660e748f61c02dc0f88ded3eebd2f27ba)) implement two custom signer loaders. One for relative path URIs, where a base bath is expected as environment variable (as you requested). And one for URIs, which point to environment variable names, from which the key can be loaded.
> 
> As you can see, they only need to be registered in the Signer API's dispatch table for `SignerStore.get` to use them. No other change is needed.
> 
> Btw. I mostly removed the config parsing code for emphasis (to show how the code becomes even simpler). If we want to stay backwards compatible we can easily revert that and make the SignerStore use the old-style config if it exists, but expect URIs on keys otherwise.

I like what you did here, but will feel more confident when we have the other part which is at the CLI.

lukpueh:
will be superseded by #451 and #452

lukpueh:
Closing in favour of #455

PR #429, opened 2023-12-10
Improve test running

KAUTH:
Improve test duration with the `pytest-xdist` plugin


KAUTH:
@kairoaraujo @MVrachev I can't add you via the reviewers so tagging you. Also. it would be nice to add this to the sprint as well, under the To-Review column.

KAUTH:
@kairoaraujo @MVrachev I fixed the test, but I'm more worried about why it had been failing or why it hadn't been failing before.

In short, `update_snapshot` function gives a type of `Metafile` (from tuf python client) to `snapshot.signed.meta` keys, so we have to assert accordingly in the test. My question is how it passed before? I don't see any recent changes in the tuf client or in the worker on how this is handled.

kairoaraujo:
> @kairoaraujo @MVrachev I fixed the test, but I'm more worried about why it had been failing or why it hadn't been failing before.
> 
> In short, `update_snapshot` function gives a type of `Metafile` (from tuf python client) to `snapshot.signed.meta` keys, so we have to assert accordingly in the test. My question is how it passed before? I don't see any recent changes in the tuf client or in the worker on how this is handle
> 

I think it was using a previous test mock return. 
While using async tests it is not ordered anymore. 

> 
> Also, somehow `coverage` was affected...

That because there is returns in tests. 
Just exclude tests/ from coverage.

MVrachev:
> That because there is returns in tests.
> Just exclude tests/ from coverage.

Yes, this should fix the issue.

KAUTH:
@MVrachev @kairoaraujo Should be ready now!

kairoaraujo (approved):

kairoaraujo (approved):

KAUTH:
@kairoaraujo some of the functional tests failed for what seems to be an external reason from GH actions, could you reapprove so that the tests are triggered?

kairoaraujo (approved):

PR #432, opened 2023-12-19
Refactor update_snapshot & _run_online_roles_bump

MVrachev:
The goal of the "update_snapshot" refactoring is to:
1. make a better separation between the two usages of "update_snapshot"
- the first is to update snapshot and all roles that were given with the "targets_roles" argument
- the second is to update snapshot and ALL bin target roles In the end, I wanted to be able to call "update_snapshot" with "bump_all" without having to pass the "targets_roles" argument.
2. fix a bug where when calling "update_snapshot" the resulting new snapshot meta didn't contain updated information for targets.json This resulted in a bug that when calling "_run_online_roles_bump", which updated targets.json, we saved the new targets.json, but the new snapshot.json didn't reference it.

The second part of this refactoring is about "_run_online_roles_bump". Again there were two goals of this refactoring:
1. make sure we call "update_snapshot" correctly. This means when we force an update of all target bins we shouldn't also pass the "targets_roles" list.
2. make it clearer how we use the "force" argument:
- when "force = True" we force a bump of all bins and targets.json
- when "force = False" we bump only those bins and targets.json which have expired.

Finally, I did some small tweaks on our tests with the idea to improve our asserts and checks.

Besides unit tests, it was tested locally with a worker instance and triggered the automatic online roles bump.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -331,6 +331,9 @@ class MetadataRepository:
                     version=bins_md.signed.version
                 )
 
+            bins = "".join(target_roles)
+            logging.info(f"Bumped all expired target 'bin' roles: {bins}")
+
         elif bump_all:
             db_target_roles = targets_crud.read_all_roles(self._db)
             for db_role in db_target_roles:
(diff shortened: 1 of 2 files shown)


MVrachev:
I am planning to use the changes of this pr when I prepare the worker changes required for: https://github.com/repository-service-tuf/repository-service-tuf-api/issues/399

MVrachev:
@kairoaraujo, the pr is ready for review.

MVrachev:
This time for real @kairoaraujo, the pr is ready for review.

kairoaraujo (approved):

PR #435, opened 2024-01-05
Add support for force online metadata update

MVrachev:
Add support for force online metadata update

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -1221,50 +1221,46 @@ class MetadataRepository:
 
         return True
 
-    def _run_force_online_metadata_update(
-        self, payload: List[str]
-    ) -> List[str]:
+    def _run_force_online_metadata_update(self, roles: List[str]) -> List[str]:
         "Run the actual metadata update for set of online roles."
         roles_diff: List[str] = []
-        delegated_roles: List[str]
-        if Roles.BINS.value in payload:
+        delegated_roles: List[str] = []
+        if Roles.BINS.value in roles:
             delegated_roles = [Roles.BINS.value]
         else:
-            delegated_roles = [r for r in payload if not Roles.is_role(r)]
+            delegated_roles = [r for r in roles if not Roles.is_role(r)]
 
-        delegated_roles_in_payload = any(r in payload for r in delegated_roles)
-        if Targets.type in payload and delegated_roles_in_payload:
+        if Targets.type in roles and len(delegated_roles) > 0:
             self._run_online_roles_bump(force=True)
             roles_diff = [
                 Targets.type, *delegated_roles, Snapshot.type, Timestamp.type
             ]
 
-        elif Targets.type in payload or delegated_roles_in_payload:
-            if Targets.type in payload:
+        elif Targets.type in roles or len(delegated_roles) > 0:
+            if Targets.type in roles:
                 targets = self._storage_backend.get(Targets.type)
                 self._bump_and_persist(targets, Targets.type)
                 self.bump_snapshot(force=True)
                 roles_diff = [Targets.type, Snapshot.type, Timestamp.type]
             else:
-                bump_all = delegated_roles == [Roles.BINS.value]
-                if bump_all:
-                    # Bump all bins
-                    self._update_timestamp(
-                        self._update_snapshot(bump_all=True)
+                target_roles = delegated_roles
+                if delegated_roles == [Roles.BINS.value]:
+                    # Set the actual names of the bins
+                    target_roles = self._settings.get_fresh(
+                        "DELEGATED_ROLES_NAMES"
                     )
-                else:
-                    # Bump custom target delegations
-                    self._update_timestamp(
-                        self._update_snapshot(target_roles=delegated_roles)
-                    )
+
+                self._update_timestamp(
+                    self._update_snapshot(target_roles=target_roles)
+                )
 
                 roles_diff = [*delegated_roles, Snapshot.type, Timestamp.type]
 
-        elif Snapshot.type in payload:
+        elif Snapshot.type in roles:
             self.bump_snapshot(force=True)
             roles_diff = [Snapshot.type, Timestamp.type]
 
-        elif Timestamp.type in payload:
+        elif Timestamp.type in roles:
             snapshot = self._storage_backend.get(Snapshot.type)
             self._update_timestamp(snapshot.signed.version)
             roles_diff = [Timestamp.type]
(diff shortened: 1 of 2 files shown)


MVrachev:
It's best to test this together with https://github.com/repository-service-tuf/repository-service-tuf-api/pull/513

MVrachev:
@kairoaraujo it's ready for a review.

MVrachev:
I invited @fsavoia and @ivanayov for a review.

repository_service_tuf_worker/repository.py
```diff
@@ -1108,6 +1112,110 @@ def bump_online_roles(self, force: Optional[bool] = False) -> bool:
...
+            )
+        if len(payload["roles"]) < 1:
+            return self._task_result(
+                TaskName.FORCE_ONLINE_METADATA_UPDATE,
+                True,
+                {
+                    "message": "No online metadata roles given",
+                },
+            )
+
```
kairoaraujo, line 1173: In the API, [we allow users to submit offline roles without checking it (example: Root)](https://github.com/repository-service-tuf/repository-service-tuf-api/blob/5594c1684064705c12f660ed68858c82636ba154/repository_service_tuf_api/metadata.py)


In the Worker, we also don't verify it. It will return a task SUCCESS with status `True`.
I suggest we verify if the role

Note:  We can verify Root and Targets now. We have in the settings/redis [a setting if the `Targets` is offline or online key](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/0463b088095b0f8a98be91b9f80591f7d8ffa088/repository_service_tuf_worker/repository.py#L493-L495) -- for future feature that allows offline targets.

kairoaraujo (changes requested):
The feature is nice. Thank you @MVrachev.
I'm adding one specific comment.

repository_service_tuf_worker/repository.py
MVrachev: I changed it and now we only allow online roles and specifically check if targets is an online role.

MVrachev:
Ready for another review.

kairoaraujo:
I'm reviewing it now

repository_service_tuf_worker/repository.py
```diff
@@ -61,7 +61,13 @@ class Roles(enum.Enum):
 
     @staticmethod
     def online_roles() -> List[str]:
-        return [Targets.type, Snapshot.type, Timestamp.type, "bins"]
+        online_roles = [Targets.type, Snapshot.type, Timestamp.type, "bins"]
+        settings_repository = get_worker_settings()
+        settings_repository.reload()
+        if not settings_repository.get_fresh("TARGETS_ONLINE_KEY", True):
+            online_roles = [Snapshot.type, Timestamp.type, "bins"]
```
ivanayov, line 68: Shouldn't bins be excluded as well?
I might be missing some details from the workflow - if so, please ignore to not waste time with the comment. 
Otherwise, given that bins are delegates and `TARGETS_ONLINE_KEY` is `False`, it looks incorrect to have them in the online roles.

repository_service_tuf_worker/repository.py
```diff
@@ -1078,8 +1085,11 @@ def bump_online_roles(self, force: Optional[bool] = False) -> bool:
                 expire (`self._hours_before_expire`)
         """
         logging.debug(f"Configured timeout: {self._timeout}")
-        if self._settings.get_fresh("BOOTSTRAP") is None:
-            logging.info("[automatic_version_bump] No bootstrap, skipping...")
+        bootstrap = self._settings.get_fresh("BOOTSTRAP")
+        if bootstrap is None or "pre-" in bootstrap or "signing-" in bootstrap:
+            logging.info(
+                "[automatic_version_bump] Bootstrap not completed, skipping..."
```
ivanayov, line 1091: "No bootstrap or not completed, skipping"?

repository_service_tuf_worker/repository.py
```diff
@@ -1108,6 +1118,104 @@ def bump_online_roles(self, force: Optional[bool] = False) -> bool:
...
+                )
+
+            status_lock_targets = True
+        except redis.exceptions.LockNotOwnedError:
+            # The LockNotOwnedError happens when the task exceeds the timeout,
+            # and another task owns the lock.
+            # If the task time out, the lock is released. If it doesn't finish
+            # properly, it will raise (fail) the task. Otherwise, the ignores
+            # the error because another task didn't lock it.
+            if status_lock_targets is False:
```
ivanayov, line 1197: It will always be `False` as the exception is thrown prior ` status_lock_targets = True`.

repository_service_tuf_worker/repository.py
MVrachev: Bins could use the online key. 
Even though targets could be offline in the future this doesn't mean we are required to use offline keys for all its delegations. The security properties coming from targets utilizing an offline key wouldn't be lost as when adding, removing or downloading artifact information all requires going through `targets` first which will be cryptographically signed by an offline key.

repository_service_tuf_worker/repository.py
MVrachev: I think `Bootstrap not completed` could be interpreted by both:
- not started at all
- not signed with threshold of keys

It's good that the user have `GET /api/v1/bootstrap` to understand that bootstrap is in the process of signing: https://github.com/repository-service-tuf/repository-service-tuf-api/blob/277d73b1610a5ec96c4af4d3ccbc8bdbfc69bcf9/repository_service_tuf_api/bootstrap.py#L128
and `GET /api/v1/metadata` to get the exact content of this bootstrap.

In conclusion, I don't we need to change the message to a longer now.

kairoaraujo (commented):
Is it better to merge custom delegated roles and rebase them?

MVrachev:
@kairoaraujo ready to test together with https://github.com/repository-service-tuf/repository-service-tuf-api/pull/513
Support for pushing a new version of custom target delegation is there.

repository_service_tuf_worker/repository.py
```diff
@@ -1208,6 +1221,116 @@ def bump_online_roles(self, force: Optional[bool] = False) -> bool:
 
         return True
 
+    def _run_force_online_metadata_update(self, roles: List[str]) -> List[str]:
```
kairoaraujo, line 1224: I'm wondering if the `metadata_update` is the best name here, as it confuses our process of updating the metadata.

kairoaraujo (approved):

repository_service_tuf_worker/repository.py
MVrachev: Well it is actually metadata update...
Hard to get a name correctly...

PR #439, opened 2024-01-21
refactoring: update Task response data structure

### LINKED ISSUE #351: Enhancement: enforce `message` and `error` in case of status `False`
https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/2e071289e5353d5f47001e35942f477f82f4074f/repository_service_tuf_worker/repository.py#L626-L634

We now have this nonwriting standard that returns a task result with a `message` and `error` when the task status is `False`.

It would be nice to enforce it while someone is writing the code to avoid relay on the review.

fsavoia:
Enhance the consistency of Task Results in our codebase.

This PR creates a consistent data structure for Task responses.

Fixes #351

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -94,6 +94,8 @@ class TaskResult:
     task: TaskName
     status: bool
     last_update: datetime
+    message: Optional[str]
+    error: Optional[str]
     details: Dict[str, Any]
 
 
@@ -553,12 +555,19 @@ class MetadataRepository:
 
     @staticmethod
     def _task_result(
-        task: TaskName, status: bool, details: Dict[str, Any]
+        task: TaskName,
+        status: bool,
+        details: Dict[str, Any],
     ) -> Dict[str, Any]:
         """Returns a standardized Task Result to the Result Backend Service"""
+        message = details.pop("message", None) if not status else None
+        error = details.pop("error", None) if not status else None
+
         result = TaskResult(
             task=task,
             status=status,
+            message=message,
+            error=error,
             details=details,
             last_update=datetime.now(),
         )
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -96,7 +96,7 @@ class TaskResult:
     last_update: datetime
     message: str
     error: Optional[str]
-    details: Dict[str, Any]
+    details: Optional[Dict[str, Any]]
 
 
 class MetadataRepository:
@@ -400,9 +400,11 @@ class MetadataRepository:
             update_state(
                 state=state,
                 meta={
-                    "published_roles": completed_roles,
-                    "roles_to_publish": f"{list(bin_targets.keys())}",
-                    "status": "Publishing",
+                    "details": {
+                        "published_roles": completed_roles,
+                        "roles_to_publish": f"{list(bin_targets.keys())}",
+                    },
+                    "message": "Publishing",
                     "last_update": datetime.now(),
                     "exc_type": exc_type,
                     "exc_message": exc_message,
@@ -558,7 +560,7 @@ class MetadataRepository:
         task: TaskName,
         message: str,
         error: Optional[str],
-        details: Dict[str, Any],
+        details: Optional[Dict[str, Any]],
     ) -> Dict[str, Any]:
         """Returns a standardized Task Result to the Result Backend Service"""
         if error is not None:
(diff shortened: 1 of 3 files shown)


kairoaraujo (changes requested):
Hi @fsavoia, thanks for your PR 🎉 

The solution you applied works, but it looks like a workaround.

Here are my suggestions for this PR:

1. Modify the `_task_result` function to receive `message` as required and `error` as Optional
2. When `error` is not None, make the `status` True
3. Modify all the `_task_result()` calls following the new implemantation
4. Modify the tests

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,20 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
 
     @staticmethod
     def _task_result(
-        task: TaskName, status: bool, details: Dict[str, Any]
+        task: TaskName,
+        status: bool,
```
kairoaraujo, line 559: I think you can remove it from the `_task_result` parameters.
If you have error not None, it means the `status` is `False`.

kairoaraujo (changes requested):

kairoaraujo (dismissed):

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,21 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
 
     @staticmethod
     def _task_result(
```
kairoaraujo, line 557:
```suggestion
    def _task_response(
```

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,21 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
...
-        task: TaskName, status: bool, details: Dict[str, Any]
+        task: TaskName,
+        message: str,
```
kairoaraujo, line 559:
```suggestion
        result: str,
```

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,21 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
...
+        error: Optional[str],
+        details: Dict[str, Any],
```
kairoaraujo, line 561:
```suggestion
        details: Optional[Dict[str, Any]],
```

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,21 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
...
     ) -> Dict[str, Any]:
         """Returns a standardized Task Result to the Result Backend Service"""
+        if error is not None:
+            status = False
+        else:
+            status = True
         result = TaskResult(
```
kairoaraujo, line 568:
```suggestion
        result = TaskData(
```

repository_service_tuf_worker/repository.py
```diff
@@ -553,12 +555,21 @@ def _bootstrap_online_roles(self, root: Metadata[Root]):
...
             task=task,
             status=status,
+            message=message,
```
kairoaraujo, line 571:
```suggestion
            result=result,
```

kairoaraujo (changes requested):
I discussed with @fsavoia this morning that we can improve the consistency of Task Results.

The first thing is the primary data structure we need to follow is the Celery Task result in case of `FAILURE` as we cannot handle it, it is returned to the backend result automatically.

This data structure is:

```json
{
  "data": {
    "task_id": "33e66671dcc84cdfa2535a1eb030104c",
    "state": "FAILURE",
    "result": "failure message -- exception failure",
  },
  "message": "Task state."
}
```

The first thing here is that `result` is a string and we currently say in the [API that is a `string` or `Dict[str, any`]](https://repository-service-tuf.github.io/repository-service-tuf-api/#/Task/get_api_v1_task__get).

https://github.com/repository-service-tuf/repository-service-tuf-api/blob/e5747ffe82f2f852eb720d87fa0f70a0d5d96a0f/repository_service_tuf_api/tasks.py#L79-L83

We want to keep it consistent, and we will make it always a string.

We will rename our `result` as  `response` with data as `Optional[Dict[str, Any]]`

The detailed changes follow in the code review.

This will require changes on the PR https://github.com/repository-service-tuf/repository-service-tuf-api/pull/518 as well :)

kairoaraujo (approved):

PR #441, opened 2024-01-22
Remove preliminary LocalKeyVault private key check

lukpueh:
Previously, `LocalKeyVault.configure` required loading at least one valid key at start-up time. This patch removes the requirement for the following reasons:

- It should be possible to start the RSTUF container without private key configuration. This makes incremental rollout easier.
- The ultimate "validity" of a private key in RSTUF can only be determined, if the related public key is known, which is not the case until the bootstrap event. Any prior checks have limited value.
- In the future, the exact private key location will be configured via the related public key. This makes it impossible to even load a private key before the related public key is known (see #580).

**Change details**

- `LocalKeyVault.configure` can now return an empty LocalKeyVault.
- Related tests are changed to expect an empty LocalKeyVault, instead of an exception.

### DIFF
--- a/repository_service_tuf_worker/services/keyvault/local.py
+++ b/repository_service_tuf_worker/services/keyvault/local.py
@@ -120,11 +120,6 @@ class LocalKeyVault(IKeyVault):
                 logging.error("Key is invalid")
                 pass
 
-        if len(parsed_keys) == 0:
-            raise KeyVaultError(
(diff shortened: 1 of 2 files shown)


lukpueh:
@kairoaraujo, what do you think about an alternative solution, where `LocalKeyVault` stays as is, but `RSTUF_KEYVAULT_BACKEND` becomes optional?  This would also allow us to start the container without keys.

PR #452, opened 2024-02-06
Add AWS KMS support to SignerStore

lukpueh:
*[EDIT 02/14/24: remove poc/draft status]* 

Add AWS KMS support to SignerStore and test with localstack (`tox -e local-aws-kms`)

**Change details**

* Whitelist required ambient settings in SignerStore (see tox.ini for required settings)

* Add independent tox environment to init/cleanup localstack, configure ambient AWS KMS credentials, create a test key, and run the test.

* Add test to "import" test public key from AWS KMS and configure private key URI - this would typically happen in a key management UI (e.g. RSTUF CLI) - and use `SignerStore.get` to load the signer.

**Todo**

This PR adds a single unit test method, which is ran against a 3rd party service. This doesn't fit into the current test architecture, where 3rd party services are consistently mocked in unit tests, and only included in more comprehensive functional tests. 

Please advise how to best test this!

[^1]: ignores [preparatory refactor commit](c0c9a4e52668fe2cc816d2925f0165ebc81fe6d1) and [ambient settings whitelist](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452/files#diff-480fc64f4facc7ed2848c406223c2fa96d1d0b5dbb672ff78b4859d638a768e0R100-R103)

PR #415, opened 2023-10-19
Add support for AWSKMS

MVrachev:
Fixes #344 

Add support for AWSKMS and do small refactorings.

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>

PR #455, opened 2024-02-12
Add FileNameSigner and register in SignerStore

lukpueh:
The custom FileNameSigner is similar to the previously registered CryptoSigner provided by securesystemslib, with the following differences:

- uri format is `fn:<file name>`
- a base path for the key in `<file name>` must be provided via environment variable `RSTUF_ONLINE_KEY_DIR`
- only **unencrypted** keys are supported

supersedes #427

### DIFF
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -234,10 +234,7 @@ service without a keyvault backend, but you need to configure one before the
 
 #### (Optional, *experimental*) `RSTUF_ONLINE_KEY_DIR`
 
-Directory path to online signing key. The setting is required, if the related
-public key metadata includes a field `x-rstuf-online-key-uri` with a value
-`fn:<file name>`. The full path is constructed by joining directory path
-and file name. The expected private key file format is unencrypted PKCS8/PEM.
+Directory path for online signing key file. Expected file format is unencrypted PKCS8/PEM.
 
 Make sure to use the secrets management service of your deployment platform to
 protect the private key!
@@ -245,21 +242,8 @@ protect the private key!
 Replaces `RSTUF_KEYVAULT_BACKEND` and related settings.
 
 Example:
--  `RSTUF_ONLINE_KEY_DIR=/run/secrets`
-- TUF root metadata (part):
-  ```
-  "keys": {
-   "c6d8bf2e4f48b41ac2ce8eca21415ca8ef68c133b47fc33df03d4070a7e1e9cc": {
-    "keytype": "ed25519",
-    "keyval": {
-     "public": "4f66dabebcf30628963786001984c0b75c175cdcf3bc4855933a2628f0cd0a0f"
-    },
-    "scheme": "ed25519",
-    "x-rstuf-online-key-uri": "fn:my_online_key"
-   }
-   ...
-  }
-- RSTUF worker expects related private key under  `/run/secrets/my_online_key`
+- `RSTUF_ONLINE_KEY_DIR=/run/secrets`
+- RSTUF worker expects related private key under  `/run/secrets/<file name>`
 
 
 
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -232,6 +232,37 @@ service without a keyvault backend, but you need to configure one before the
 
   Example: ``RSTUF_LOCAL_KEYVAULT_KEYS=/run/secrets/ONLINE_KEY_1:/run/secrets/ONLINE_KEY_2``
 
+#### (Optional, *experimental*) `RSTUF_ONLINE_KEY_DIR`
+
+Directory path to online signing key. The setting is required, if the related
+public key metadata includes a field `x-rstuf-online-key-uri` with a value
+`fn:<file name>`. The full path is constructed by joining directory path
+and file name. The expected private key file format is unencrypted PKCS8/PEM.
+
+Make sure to use the secrets management service of your deployment platform to
+protect the private key!
+
+Replaces `RSTUF_KEYVAULT_BACKEND` and related settings.
+
+Example:
+-  `RSTUF_ONLINE_KEY_DIR=/run/secrets`
+- TUF root metadata (part):
+  ```
+  "keys": {
+   "c6d8bf2e4f48b41ac2ce8eca21415ca8ef68c133b47fc33df03d4070a7e1e9cc": {
+    "keytype": "ed25519",
+    "keyval": {
+     "public": "4f66dabebcf30628963786001984c0b75c175cdcf3bc4855933a2628f0cd0a0f"
+    },
+    "scheme": "ed25519",
+    "x-rstuf-online-key-uri": "fn:my_online_key"
+   }
+   ...
+  }
+- RSTUF worker expects related private key under  `/run/secrets/my_online_key`
+
+
+
 
 #### (Optional) `RSTUF_WORKER_ID`
 
--- a/repository_service_tuf_worker/signer.py
+++ b/repository_service_tuf_worker/signer.py
@@ -24,7 +24,7 @@ class FileNameSigner(CryptoSigner):
 
     Provide method to load **unencrypted** PKCS8/PEM private key from file.
 
-    File path is constructed by joining base path in envrionment variable
+    File path is constructed by joining base path in environment variable
     ``RSTUF_ONLINE_KEY_DIR`` with file in ``priv_key_uri``.
 
     NOTE: Make sure to use the secrets management service of your deployment
@@ -40,7 +40,7 @@ class FileNameSigner(CryptoSigner):
     Raises:
         KeyError: RSTUF_ONLINE_KEY_DIR environment variable not set
         OSError: file cannot be loaded
-        ValueError: uri has no filenmae, or private key cannot be decoded,
+        ValueError: uri has no file name, or private key cannot be decoded,
                 or type does not match public key
         cryptography.exceptions.UnsupportedAlgorithm: key type not supported
 
--- a/repository_service_tuf_worker/signer.py
+++ b/repository_service_tuf_worker/signer.py
@@ -1,22 +1,83 @@
 # SPDX-FileCopyrightText: 2024 Repository Service for TUF Contributors
 #
 # SPDX-License-Identifier: MIT
+import os
+from pathlib import Path
+from typing import Optional
 
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
 from dynaconf import Dynaconf
 from securesystemslib.signer import (
     SIGNER_FOR_URI_SCHEME,
     CryptoSigner,
     Key,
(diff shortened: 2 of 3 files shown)


repository_service_tuf_worker/signer.py
```diff
@@ -1,22 +1,83 @@
...
+        secrets_handler: Optional[SecretsHandler] = None,
+    ) -> "FileNameSigner":
+        _, _, file_name = priv_key_uri.partition(":")
+        if not file_name:
+            raise ValueError(
+                f"bad uri: expected '{cls.SCHEME}:<file name>', "
+                f"'got {priv_key_uri}'"
+            )
+
+        dir_ = os.environ[cls.DIR_VAR]
```
kairoaraujo, line 65: RSTUF stores all `RSTUF_*` environment variables in Dynaconf as settings.
It is used by the `SignerStore`, but cannot be used in the custom signer `FileNameSigner` will use it outside the RSTUF instance.

Is there a way to keep it centralized in the RSTUF dynaconf settings?

repository_service_tuf_worker/signer.py
```diff
@@ -1,22 +1,83 @@
...
+        KeyError: RSTUF_ONLINE_KEY_DIR environment variable not set
+        OSError: file cannot be loaded
+        ValueError: uri has no filenmae, or private key cannot be decoded,
+                or type does not match public key
+        cryptography.exceptions.UnsupportedAlgorithm: key type not supported
+
+    """
+
+    SCHEME = "fn"
+    DIR_VAR = "RSTUF_ONLINE_KEY_DIR"
```
kairoaraujo, line 49: I think the settings (env variable) can be more specific as it is not for all online key schemas, only for  `fn`
`RSTUF_ONLINE_KEY_FN_DIR`.

kairoaraujo (commented):
Creating a custom signer is a good approach, as the sslib API makes it very easy to implement.
The only thing we can discuss is that it will use a setting outside of RSTUF settings (Dynaconf) as I commented below.

kairoaraujo:
> Creating a custom signer is a good approach, as the sslib API makes it very easy to implement. The only thing we can discuss is that it will use a setting outside of RSTUF settings (Dynaconf) as I commented below

I was wondering if isn't better to implement the logic in the `SignerStore`, in the `get`. 

The user from the RSTUF container will have the parameter `RSTUF_ONLINE_KEY_FILE_DIR` as optional.
If the user sets it and the key is scheme file, it appends as suffix in value.

Advantages
- The CLI user doesn’t need to choose between File and FileName (which can be confused)
- It become simple and the user in the CLI can give or not the dir suffix
- We don’t need to maintain also a custom signer in the CLI and Worker
- We keep the RSTUF Worker settings with Dynaconf

repository_service_tuf_worker/signer.py
lukpueh: I'm fine either way, although I think that `RSTUF_ONLINE_KEY_DIR` seems pretty self-explanatory - directory for rstuf online key - and adding an `fn` infix doesn't make it clearer but more confusing.

I assume you want to bind the name to the uri scheme or to FileNameSigner? Ideally, the user doesn't need to know about that binding (I'll talk more about it below).

lukpueh:
> I was wondering if isn't better to implement the logic in the `SignerStore`, in the `get`.

I strongly recommend to not do that. The main appeal of the Signer API, besides the `sign` interface, is that signer access/loading is an implementation detail of the individual signer implementation. This interface allowed me to add AWS support with 0 lines of code changes in SignerStore (#237), which is substantially less than a prior wip attempt that implemented the loading logic in the worker (#415).

Case handling different signers in `SignerStore` weakens the universal `Signer.from_priv_key_uri` interface of passing specific information about the signer via public key and uri, and providing additional information ambiently (e.g. via env var).

> The user from the RSTUF container will have the parameter `RSTUF_ONLINE_KEY_FILE_DIR` as optional. If the user sets it and the key is scheme file, it appends as suffix in value.
> 
> Advantages
> 
> * The CLI user doesn’t need to choose between File and FileName (which can be confused)
> * It become simple and the user in the CLI can give or not the dir suffix

The cli user doesn't need to know about `File` or `FileName` at all. IIRC we agreed that the cli only asks the user if they want to provide the online key as file, then transparently adds `fn:<keyid>` uri to the public key, and informs the user to:
- set a key base path via `RSTUF_ONLINE_KEY_DIR` env var
- provide the key in `RSTUF_ONLINE_KEY_DIR / <keyid>`

I really don't think that the cli should also provide an option to add a `file:/abs/path/to/<keyid>`. If we do want to support absolute file path uris, we can easily change `FileNameSigner.from_priv_key_uri` to threat the file name in the uri as absolute path if  `RSTUF_ONLINE_KEY_DIR` is not set.

But let's not get too caught up with file-based online keys. Given that the Signer API already supports multiple cloud kms services, file-based keys should be an edge case mainly for development.

> * We don’t need to maintain also a custom signer in the CLI and Worker

 FileNameSigner is only needed in the worker and it comes at about ~20 lines of self-contained code with a clear interface. IIUC, you suggested above to move that code into the signer store, so I don't think maintaining this here is an issue. That said, @jku thinks that there is value in upstreaming the FileNameSigner (https://github.com/repository-service-tuf/repository-service-tuf/issues/580#issuecomment-1913108310)

> * We keep the RSTUF Worker settings with Dynaconf

This is a legitimate concern and we need to find a solution that also works for other signers with ambient configuration, which require translation from the RSTUF namespace to the domain namespace, e.g. where `RSTUF_AWS_ACCESS_KEY_ID` becomes [`AWS_ACCESS_KEY_ID`](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/452/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R79-R82).

I think `SignerStore` could do that, e.g.:

```python

class SignerStore:
    def __init__(self, settings: Dynaconf):
        ...
        self._env: dict[str, str] = {}
        for k, v in settings:
            if k.startswith("RSTUF_ONLINE_KEY_"):
                self._env[TRANSLATE_ENV(k)] = v

    def get(self, key: Key) -> Signer:
        ...
        with SET_ENV(self._env):
            signer = Signer.from_priv_key_uri(uri, key)
```

repository_service_tuf_worker/signer.py
lukpueh: See code snippet in [comment below](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/455#issuecomment-1940942231).

kairoaraujo:
Thank you for all the detailed information, @lukpueh

kairoaraujo:
> ```python
> class SignerStore:
>     def __init__(self, settings: Dynaconf):
>         ...
>         self._env: dict[str, str] = {}
>         for k, v in settings:
>             if k.startswith("RSTUF_ONLINE_KEY_"):
>                 self._env[TRANSLATE_ENV(k)] = v
> 
>     def get(self, key: Key) -> Signer:
>         ...
>         with SET_ENV(self._env):
>             signer = Signer.from_priv_key_uri(uri, key)
> ```

The variable in settings will be `setttings.ONLINE_KEY_DIR` (accessible also by `settings.get(ONLINE_KEY_DIR)`).
Dynaconf removes the prefix `RSTUF_` to identify specific settings in the container environment.

Please, can you include the `RSTUF_ONLINE_KEY_DIR` in the `docs/source/guide/Docker_README` as well?

lukpueh:
@kairoaraujo, I added the requested changes. Let me know what you think.

repository_service_tuf_worker/signer.py
```diff
@@ -1,51 +1,140 @@
...
+        - additional signer settings can be provided "ambiently" (see __init__)
         - RSTUF_KEYVAULT_BACKEND is used as fallback, if no URI is included
 
         """
 
         if key.keyid not in self._signers:
             if uri := key.unrecognized_fields.get(RSTUF_ONLINE_KEY_URI_FIELD):
-                self._signers[key.keyid] = Signer.from_priv_key_uri(uri, key)
+                # (Re-)export ambient settings in isolated environment
+                with isolated_env(self._ambient_settings):
```
lukpueh, line 127: I wonder if we really need an isolated environment. I added it mostly to avoid weird conflicts between the `RSTUF_` prefixed config and the signer-specific expected environment variables.

Let's take AWS as example. The AWSSigner implementation in securesystemslib (used by #452) expects an environment variable `AWS_ENDPOINT_URL`. But we would likely ask the user to provide that value via `RSTUF_AWS_ENDPOINT_URL` setting.

Now, how should SignerStore behave if the `RSTUF_AWS_ENDPOINT_URL` setting is missing, but `AWS_ENDPOINT_URL` is defined? Unless we isolate (or vet) the environment, AWSSigner will happily accept the environment variable.

lukpueh: A limitation of the isolated environment is that the signer can only access the ambient configuration when it is created (`from_priv_key_uri`) but not when it is used (`sign`).

This is not an issue for `FileNameSigner` or `AWSSigner` (tested in https://github.com/repository-service-tuf/repository-service-tuf-worker/commit/e1514aac496a5051afece6cad24fc3065bfedea5), but may be one for other Signer implementations.

lukpueh: I suggest we move forward with the isolated environment and re-consider, if we need a Signer implementation where the does not work.

docs/source/guide/Docker_README.md
```diff
@@ -232,6 +232,37 @@ service without a keyvault backend, but you need to configure one before the
 
   Example: ``RSTUF_LOCAL_KEYVAULT_KEYS=/run/secrets/ONLINE_KEY_1:/run/secrets/ONLINE_KEY_2``
 
+#### (Optional, *experimental*) `RSTUF_ONLINE_KEY_DIR`
+
+Directory path to online signing key. The setting is required, if the related
+public key metadata includes a field `x-rstuf-online-key-uri` with a value
+`fn:<file name>`. The full path is constructed by joining directory path
+and file name. The expected private key file format is unencrypted PKCS8/PEM.
```
kairoaraujo, line 240: nit: focus on the deployment without need deep info on the TUF metadata structure or Signer API

```suggestion
Directory path to online signing key. The setting is required if using provide the online key as file. The expected private key file format is unencrypted PKCS8/PEM.
```

kairoaraujo (dismissed):

docs/source/guide/Docker_README.md
lukpueh: > nit: focus on the deployment without need deep info on the TUF metadata structure or Signer API

Makes sense. 

I think there is a stray "provide" in your suggestion. What about just
 
```suggestion
Directory path for online signing key file. Expected private format is unencrypted PKCS8/PEM.
```

kairoaraujo (approved):

PR #457, opened 2024-02-15
Add support for custom delegated targets

### LINKED ISSUE #438: Task: add support for custom delegated Target roles
### What is the task about?

We want to add support for custom delegated Target roles.
To achieve this we should expect an additional optional argument called `custom_target_roles` inside the bootstrap ceremony payload. This argument is defined in https://github.com/repository-service-tuf/repository-service-tuf-api/issues/517
We need to:
1. check if this argument was passed
2. parse it
3. create custom target role metadata files from the given information
4. save the custom target roles information in our database using the CRUD operations we use for hash bin delegation roles
5. save the expiration setting for each custom delegated role. 
I propose using a common name for this expiration setting for all custom target delegated roles: `CUSTOM_TARGET_<TARGET_NAME>_EXPIRATION`

The second change that is required is to make sure that all of the custom target delegation roles will be automatically updated when calling `bump_online_roles` every 10 minutes: https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/486e98cdbc16dd41c62c0ec8b473e2f94312ee9c/app.py#L139

TODO:
- [x] update `bootstrap` ceremony
- [ ] update `bump_online_roles`
- [ ] update `update_config` to update custom target roles expiration

### Parent feature

[Feature: Create/Remove custom delegate Target Roles](https://github.com/repository-service-tuf/repository-service-tuf/issues/354)

### References

- API: https://github.com/repository-service-tuf/repository-service-tuf-api/issues/517

### Code of Conduct

MVrachev:
Fixes #438 

Add support for custom delegated targets which could be used by PyPI and other of our users.
This will allow for greater flexibility and we could support special use cases for bigger users.

Note: it's related to pr https://github.com/repository-service-tuf/repository-service-tuf-api/pull/542. When testing use both this and the API pr.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -371,15 +371,18 @@ class MetadataRepository:
 
         return snapshot.signed.version
 
-    def _get_path_succinct_role(self, target_path: str) -> str:
+    def _get_role_for_target_path(self, target_path: str) -> str:
         """
         Return role name by target file path
         """
-        bin_role: Metadata[Targets] = self._storage_backend.get(Targets.type)
-        bin_succinct_roles = bin_role.signed.delegations.succinct_roles
-        bins_name = bin_succinct_roles.get_role_for_target(target_path)
-
-        return bins_name
+        targets: Metadata[Targets] = self._storage_backend.get(Targets.type)
+        delegations = targets.signed.delegations
+        # Note: there could be more than one matching custom delegated role,
+        # for this target path, but we get the first match. If we want to get
+        # the most specific one we need to use _preorder_depth_first_walk
+        # of the ngclient Updater class in python-tuf.
+        role_name, _ = next(delegations.get_roles_for_target(target_path))
+        return role_name
 
     def _update_task(
         self,
@@ -897,7 +900,7 @@ class MetadataRepository:
         # This will be used to by `_update_task` for updating task status.
         bin_targets: Dict[str, List[targets_models.RSTUFTargetFiles]] = {}
         for target in targets:
-            bins_name = self._get_path_succinct_role(target["path"])
+            delegated_role = self._get_role_for_target_path(target["path"])
             db_target_file = targets_crud.read_file_by_path(
                 self._db, target.get("path")
             )
@@ -911,7 +914,7 @@ class MetadataRepository:
                         action=targets_schema.TargetAction.ADD,
                     ),
                     target_role=targets_crud.read_role_by_rolename(
-                        self._db, bins_name
+                        self._db, delegated_role
                     ),
                 )
             else:
@@ -922,10 +925,10 @@ class MetadataRepository:
                     target.get("info"),
                 )
 
-            if bins_name not in bin_targets:
-                bin_targets[bins_name] = []
+            if delegated_role not in bin_targets:
+                bin_targets[delegated_role] = []
 
-            bin_targets[bins_name].append(db_target_file)
+            bin_targets[delegated_role].append(db_target_file)
 
         # If publish_targets doesn't exists it will be True by default.
         publish_targets = payload.get("publish_targets", True)
@@ -982,7 +985,7 @@ class MetadataRepository:
         # This will be used to by `publish_targets`
         bin_targets: Dict[str, List[targets_models.RSTUFTargetFiles]] = {}
         for target in targets:
-            bins_name = self._get_path_succinct_role(target)
+            delegated_role = self._get_role_for_target_path(target)
             db_target = targets_crud.read_file_by_path(self._db, target)
             if db_target is None or (
                 db_target.action == targets_schema.TargetAction.REMOVE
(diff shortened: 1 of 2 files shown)


MVrachev:
Ready for review @kairoaraujo.

MVrachev:
There is only one change remaining in the scope of this pr - adding and removing artifacts.

MVrachev:
The pr is ready for review.
I added support for:
- bootstrap with custom target delegations
- automatic updates of custom target delegations
- update of expiration of custom target delegations
- add and remove artifacts from/to custom target delegations

So, @kairoaraujo the pr should be ready for a review.

kairoaraujo:
> The pr is ready for review. I added support for:
> 
> * bootstrap with custom target delegations
> * automatic updates of custom target delegations
> * update of expiration of custom target delegations
> * add and remove artifacts from/to custom target delegations
> 
> So, @kairoaraujo I think I implemented everything I needed.

Hi @MVrachev,
It is a fascinating feature. Thank you for your work on it.
I'm reviewing and testing it. It is impressive work!

kairoaraujo (dismissed):
Thank you for the great work.

I'm approving as I have two minor comments
- nit, just a Pythonic name.
- overall question for the function design :)

repository_service_tuf_worker/repository.py
```diff
@@ -518,53 +528,93 @@ def save_settings(self, root: Metadata[Root], settings: Dict[str, Any]):
...
+                self._bump_expiry(bins_role, BINS)
+                self._sign(bins_role)
+                self._persist(bins_role, delegated_name)
+
+    def _get_delegation_roles(
+        self, targets: Metadata[Targets]
+    ) -> Iterator[str]:
+        """Get all Targets delegation roles no matter if bins or custom."""
+        if targets.signed.delegations is None:
+            raise ValueError("Targets must have delegation, internal error")
```
MVrachev, line 578: Mention in this docstring that we raise `ValueError` exception.

repository_service_tuf_worker/repository.py
```diff
@@ -831,11 +903,11 @@ def add_targets(
             )
         # The task id will be used by `_send_publish_targets_task` (sub-task).
         task_id = payload.get("task_id")
-        # Group target files by responsible 'bins' delegated roles.
+        # Group target files by responsible delegated role.
         # This will be used to by `_update_task` for updating task status.
         bin_targets: Dict[str, List[targets_models.RSTUFTargetFiles]] = {}
         for target in targets:
-            bins_name = self._get_path_succinct_role(target["path"])
+            delegated_role = self._get_role_for_target_path(target["path"])
```
kairoaraujo, line 910: I was testing it, and if you experiment to add a payload with two targets and one of the targets has an invalid target path, it will fail.

```json
{
   "targets":[
      {
         "info":{
            "length":2004,
            "hashes":{
               "sha256":"38744fd5fa3036ad76d4539f98bcef77780bc75216c46e4de9549e90ecdf81f9"
            }
         },
         "path":"files/file3.txt"
      },
      {
         "info":{
            "length":2004,
            "hashes":{
               "sha256":"38744fd5fa3036ad76d4539f98bcef77780bc75216c46e4de9549e90ecdf81f9"
            }
         },
         "path":"file1.txt"
      }
   ]
}
```

The task will fully fail, even only the first one is good.
The problem occurs because `self._get_role_for_target_path` can raise a `StopIteration` (on `next`).

One idea add the same approach as we do in `remove_targets`.
Where we can map artifacts without paths and return in the task result.

As a workaround for this issue, I created a delegated role that has a `delegated_roles` with `path_patterns` as `*`.

repository_service_tuf_worker/repository.py
```diff
@@ -368,15 +375,18 @@ def _update_snapshot(
...
-        bins_name = bin_succinct_roles.get_role_for_target(target_path)
-
-        return bins_name
+        targets: Metadata[Targets] = self._storage_backend.get(Targets.type)
+        delegations = targets.signed.delegations
+        # Note: there could be more than one matching custom delegated role,
+        # for this target path, but we get the first match. If we want to get
+        # the most specific one we need to use _preorder_depth_first_walk
+        # of the ngclient Updater class in python-tuf.
+        role_name, _ = next(delegations.get_roles_for_target(target_path))
```
kairoaraujo, line 388: See my comment below
```suggestion
        try:
            role_name, _ = next(delegations.get_roles_for_target(target_path))
        except StopIteration:
            return None
```

repository_service_tuf_worker/repository.py
```diff
@@ -368,15 +375,18 @@ def _update_snapshot(
 
         return snapshot.signed.version
 
-    def _get_path_succinct_role(self, target_path: str) -> str:
+    def _get_role_for_target_path(self, target_path: str) -> str:
```
kairoaraujo, line 378: See my comment below
```suggestion
    def _get_role_for_target_path(self, target_path: str) -> Optional[str]:
```

kairoaraujo (changes requested):

MVrachev:
@kairoaraujo I rebased on top of main.
My changes are only at the last commit where instead of raising a `StopIteration` I return `None` as you suggested at `_get_role_for_target_path` and I logged the target without a delegated role for `add_targets` as well.

kairoaraujo (approved):

kairoaraujo (approved):

PR #475, opened 2024-03-25
Replace all mentions of "targets" with "artifacts"

MVrachev:
We want to use "targets" only in three cases:
- the top-level metadata role
- custom target delegation or targets bin role(s)
- when we mention "target files"

This is necessary as currently, we make mistakes in our code which results in messages to the user referring to "artifacts" as "targets".

Signed-off-by: Martin Vrachev [martin.vrachev@broadcom.com](mailto:martin.vrachev@broadcom.com)


kairoaraujo (approved):

kairoaraujo:
Merging as breaking changes

PR #481, opened 2024-04-04
Force UTC timezone for all datetime objects

MVrachev:
The specification defines `datetime` as UTC time zone:
https://theupdateframework.github.io/specification/latest/#file-formats-date-time
The python-tuf project updated all their datetime objects,  so it's time we do it now as well.
I updated everywhere to replace `datetime.utc()` to `datetime.now(timezone.utc)`
as `datetime.utc()`  is documented for
deprecation and is officially deprecated from python 12:
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow
Also, I added to all `datetime.now()` calls the argument `timezone.utc` to make sure that
we are consistent with python-tuf as python-tuf did that change as well: https://github.com/theupdateframework/python-tuf/pull/2573

### DIFF
--- a/repository_service_tuf_worker/models/targets/crud.py
+++ b/repository_service_tuf_worker/models/targets/crud.py
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: 2022-2023 VMware Inc
 #
 # SPDX-License-Identifier: MIT
-from datetime import datetime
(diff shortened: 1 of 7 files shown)


kairoaraujo (approved):
FT will fail due breaking compatilibity

PR #493, opened 2024-04-18
Add case for local running of FT tests from Makefile

KAUTH:
If we are not running Functional Tests on GH
actions we don't run the tests in different groups.

Related to https://github.com/repository-service-tuf/repository-service-tuf/pull/684

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,15 @@ ifeq ($(GITHUB_ACTION),)
 endif
 	docker compose run --env UMBRELLA_PATH=rstuf-umbrella --entrypoint 'bash tests/functional/scripts/run-ft-das.sh $(CLI_VERSION) $(PYTEST_GROUP) $(SLOW)' --rm repository-service-tuf-worker
 
+ft-das-local:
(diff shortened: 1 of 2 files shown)


kairoaraujo (approved):

MVrachev (approved):

KAUTH:
@MVrachev, @kairoaraujo I don't think we need to do anything else to run the FT locally in the rest of the repositories under the umbrella.
The [code](https://github.com/repository-service-tuf/repository-service-tuf/pull/684/files) was added to `repository-service-tuf` to identify if the tests are running locally. We only needed to add a [specific case](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/493/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) for the `repository-service-tuf-worker` which runs the FT from scripts inside the same repository (whereas the API and CLI use the umbrella scripts, e.g., [`rstuf-umbrella/tests/functional/scripts/run-ft-das.sh`](https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/main/Makefile#L47)).

PR #501, opened 2024-04-29
Fix two bugs & do a small refactoring

MVrachev:
In this pr I address two bugs and one refactoring:

1. Don't persist root if the online roles bump fails. This makes sure that the root is stored last when all online roles are bumped.
2. Give a `trusted_root_update()` with a more telling name. As we don't update trusted root with `_trusted_root_update()` and this makes the function name misleading.
3.  Fix: update the delegation key when bumping targets. Strangely, we forgot to change the key referenced in top level targets role when bumping targets with an online key change. This caused a bug that was reported: https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/491

Fixes #491

PR #504, opened 2024-04-30
Fix force md update with custom delegations

### LINKED ISSUE #503: Bug: force metadata update with custom delegation will update all custom delegations
### What steps did you take?

Steps to reproduce:
1. Run `make run-dev`
2. Bootstrap using custom delegation payload - https://github.com/repository-service-tuf/repository-service-tuf-api/blob/main/tests/data_examples/bootstrap/payload_custom_targets.json
3. Call `POST /api/v1/metadata/online/` with payload:
```
{
  "roles": [
    "targets",
    "foo"
  ]
}
```
4. Check that both `foo` and `bar` will be updated when only `foo` was supposed to be updated
5. Check logs and see that they say that top-level online roles and `foo` were updated. `bar` is not mentioned even though it's updated.

Update only roles that are required - `snapshot`, `timestamp`, `targets` and `foo` and DO NOT update `bar`.

### Code of Conduct

MVrachev:
Fixes #503 

Fix a bug which happened when we wanted to bump targets and a custom
delegated role at the same time which caused a bump of the target role
and all custom delegated roles at the same time.

**Note**: Merge #501 first.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -1282,16 +1282,12 @@ class MetadataRepository:
                 self._bump_and_persist(targets, Targets.type)
                 roles_diff += [Targets.type]
 
-            target_roles = delegated_roles
             if delegated_roles == [Roles.BINS.value]:
-                # Set the actual names of the bins
-                target_roles = self._settings.get_fresh(
-                    "DELEGATED_ROLES_NAMES"
+                self._update_timestamp(self._update_snapshot(bump_all=True))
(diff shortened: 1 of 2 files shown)


MVrachev:
Ready for review @kairoaraujo.
I rebased on top of main and tested it manually.

MVrachev:
@kairoaraujo I addressed your comment in my latest commit.
Ready for review.

kairoaraujo (approved):

PR #592, opened 2024-09-07
feat: Custom Delegation(Target) Roles

kairoaraujo:
It starts with a simple fix in the Database Schema:

- Include rolename as unique to avoid duplications
- Include column to set Target Role as active
- Implement several CRUD to handle inactive roles

Implements the Custom Delegation Roles
- Add support to custom delegation roles during bootstrap
- Add support for using offline and online keys in the custom delegations roles
- Add support for adding extra delegation roles afterwards
- Add support for updating delegation roles
- Add support for deleting delegation roles

TODO:
- [ ]  Tests: It just fixes errors for the tests, it requires a re-implementation of tests.

NOTE: We will prioritize the functions tests (FT) to this feature which prevent regrations or breaking the compatibility with BINS.
We will implement unit tests as improvement for development process.

See: FT propose: https://github.com/repository-service-tuf/repository-service-tuf/pull/817


kairoaraujo:
Suppressed by #593 (branch in the upstream, to show the FT)

PR #593, opened 2024-09-08
feat: Custom Delegation(Target) Roles

kairoaraujo:
It starts with a simple fix in the Database Schema:

- Include rolename as unique to avoid duplications
- Include column to set Target Role as active
- Implement several CRUD to handle inactive roles

Implements the Custom Delegation Roles
- Add support to custom delegation roles during bootstrap
- Add support for using offline and online keys in the custom delegations roles
- Add support for adding extra delegation roles afterwards
- Add support for updating delegation roles
- Add support for deleting delegation roles

TODO:
- [ ]  Unit Tests: It just fixes errors for the tests, it requires a re-implementation of tests.

### NOTE:

We will prioritize the functions tests (FT) for this feature, preventing regrations or breaking compatibility with BINS.
We will implement unit tests to improve the development process.

[Functional Tests without breaking BINS deployment.](https://github.com/repository-service-tuf/repository-service-tuf-worker/actions/runs/10756018985)

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -736,7 +736,7 @@ class MetadataRepository:
             if key not in other_keys:
                 logging.debug(f"removing key id {key}")
                 targets.signed.delegations.keys.delete(key)
-             else:
+            else:
                 logging.debug(f"key {key} used by other role")
 
     def _add_delegated_role_keys(
@@ -899,7 +899,8 @@ class MetadataRepository:
 
         if targets.signed.delegations.succinct_roles:
             raise RepositoryError(
-                "Delegations already using hash-bins, cannot delete custom roles"
+                "Delegations already using hash-bins, cannot delete custom "
+                "roles"
             )
 
         # remove role from Targets delegations and mapping role unique keys
@@ -1034,6 +1035,9 @@ class MetadataRepository:
             logging.debug("verifying delegations threshold")
             for role in success:
                 if self._validate_threshold(success[role], targets, role):
+                    # If the role uses online key it was signed by while
+                    # adding the role by self._add_metadata_delegation.
+                    # if the threshold is met, role is full signed
                     logging.debug(f"role '{role}' uses only online key")
                     snapshot.signed.meta[f"{role}.json"] = MetaFile(
                         success[role].signed.version
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -947,7 +947,8 @@ class MetadataRepository:
 
         if targets.signed.delegations.succinct_roles:
             raise RepositoryError(
-                "Delegations already using hash-bins, cannot update custom roles"
+                "Delegations already using hash-bins, cannot update"
+                "custom roles"
             )
 
         success = {}
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -383,7 +383,7 @@ class MetadataRepository:
             elif bump_all:
                 db_target_roles = targets_crud.read_all_roles(self._db)
             else:
-                ValueError("'bump_all' or 'target_roles")
+                raise ValueError("'bump_all' or 'target_roles")
 
             snapshot_meta_updated = False
             for db_role in db_target_roles:
@@ -497,7 +497,7 @@ class MetadataRepository:
 
             return snapshot.signed.version
         else:
-            None
+            return None
 
     def _update_targets_delegations_key(self, targets: Metadata[Targets]):
         """
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: MIT
 
+import copy
 import enum
 import logging
 import time
@@ -10,7 +11,7 @@ import warnings
 from dataclasses import asdict, dataclass
 from datetime import datetime, timedelta, timezone
 from math import log
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Tuple
 from urllib.parse import urlparse
 
 import redis
@@ -18,7 +19,7 @@ from celery.app.task import Task
 from celery.exceptions import ChordError
 from celery.result import AsyncResult, states
 from dynaconf.loaders import redis_loader
-from securesystemslib.exceptions import UnverifiedSignatureError
+from securesystemslib.exceptions import StorageError, UnverifiedSignatureError
 from securesystemslib.signer import (
     KEY_FOR_TYPE_AND_SCHEME,
     Key,
@@ -103,6 +104,7 @@ class TaskName(str, enum.Enum):
     PUBLISH_ARTIFACTS = "publish_artifacts"
     FORCE_ONLINE_METADATA_UPDATE = "force_online_metadata_update"
     METADATA_UPDATE = "metadata_update"
+    METADATA_DELEGATION = "metadata_delegation"
     SIGN_METADATA = "sign_metadata"
     DELETE_SIGN_METADATA = "delete_sign_metadata"
 
@@ -273,6 +275,11 @@ class MetadataRepository:
             if role_name == Root.type:
                 self.write_repository_settings("TRUSTED_ROOT", role.to_dict())
 
+            if role_name == Targets.type:
+                self.write_repository_settings(
+                    "TRUSTED_TARGETS", role.to_dict()
+                )
+
         bytes_data = role.to_bytes(JSONSerializer())
         self._storage_backend.put(bytes_data, filename)
         logging.debug(f"{filename} saved")
@@ -308,29 +315,42 @@ class MetadataRepository:
         if persist:
             self._persist(role, role_name)
 
-    def _update_timestamp(self, snapshot_version: int) -> Metadata[Timestamp]:
+    def _update_timestamp(
+        self,
+        snapshot_version: Optional[int] = None,
+        skip: Optional[bool] = False,
+    ) -> Metadata[Timestamp]:
         """
         Loads 'timestamp', updates meta info about passed 'snapshot'
         metadata, bumps version and expiration, signs and persists.
 
         Args:
-            snapshot_version: snapshot version to add to new timestamp.
-            db_targets: RSTUFTarget DB objects will be changed as published in
-                the DB SQL.
+            snapshot_version: Optional snapshot version to new timestamp.
+            skip: Skip bumping version and expiration if snpashot
+            version is not provided
         """
         timestamp: Metadata[Timestamp] = self._storage_backend.get(
             Timestamp.type, None
         )
+        if not snapshot_version and skip:
+            logging.debug(
+                "No snapshot version and skip is True. Not bumping Timestamp"
+            )
+            return timestamp
+
         timestamp.signed.snapshot_meta = MetaFile(version=snapshot_version)
 
         self._bump_and_persist(timestamp, Timestamp.type)
 
+        logging.debug("Bumped version of 'Target' role")
         return timestamp
 
     def _update_snapshot(
         self,
         target_roles: Optional[List[str]] = None,
         bump_all: Optional[bool] = False,
+        only_target: Optional[bool] = False,
+        only_snapshot: Optional[bool] = False,
     ) -> int:
         """
         Loads 'snapshot', updates meta info when 'target_roles' role names are
@@ -342,6 +362,9 @@ class MetadataRepository:
                 will NOT be taken into account.
             bump_all: Wheter to bump all delegated target roles. If provided,
                 then 'target_roles' arg is NOT taken into acount.
+            only_target: Updates only Targets role in the Snapshot Meta.
+                It doesn't Bump and Persist Targets.
+            only_snapshot: Bump and Persist Snapshot Role only.
         """
         snapshot: Metadata[Snapshot] = self._storage_backend.get(Snapshot.type)
         targets: Metadata[Targets] = self._storage_backend.get(Targets.type)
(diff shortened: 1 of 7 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,383 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+            if not_delete == delegated:
+                continue
+            other_keys += targets.signed.delegations.roles[not_delete].keyids
+
+        role_keys = targets.signed.delegations.roles[delegated.name].keyids
+        for key in role_keys:
+            if key not in other_keys:
+                logging.debug(f"removing key id {key}")
+                targets.signed.delegations.keys.delete(key)
+            logging.debug(f"key {key} used by other role")
```
MVrachev, line 739: This log should be in an else otherwise it's confusing.
```suggestion
            if key not in other_keys:
                logging.debug(f"removing key id {key}")
                targets.signed.delegations.keys.delete(key)
             else:
                logging.debug(f"key {key} used by other role")
```

repository_service_tuf_worker/repository.py
```diff
@@ -374,59 +419,85 @@ def _update_snapshot(
...
+            logging.debug("Bumped version of 'Snapshot' role")
 
-        # update expiry, bump version and persist to the storage
-        self._bump_and_persist(snapshot, Snapshot.type)
-        logging.info("Bumped version of 'Snapshot' role")
-
-        return snapshot.signed.version
+            return snapshot.signed.version
+        else:
+            None
```
MVrachev, line 500:
```suggestion
            return None
```

repository_service_tuf_worker/repository.py
```diff
@@ -342,24 +362,49 @@ def _update_snapshot(
...
-            )
+        if target_roles or bump_all:
+            if target_roles:
+                db_target_roles = targets_crud.read_roles_joint_files(
+                    self._db, target_roles
+                )
+            elif bump_all:
+                db_target_roles = targets_crud.read_all_roles(self._db)
+            else:
+                ValueError("'bump_all' or 'target_roles")
```
MVrachev, line 386:
```suggestion
                raise ValueError("'bump_all' or 'target_roles")
```

repository_service_tuf_worker/repository.py
```diff
@@ -308,29 +315,42 @@ def _bump_and_persist(
...
+            return timestamp
+
         timestamp.signed.snapshot_meta = MetaFile(version=snapshot_version)
 
         self._bump_and_persist(timestamp, Timestamp.type)
 
+        logging.debug("Bumped version of 'Target' role")
         return timestamp
 
     def _update_snapshot(
```
MVrachev, line 348: Honestly, this function isss really hard to read...
If you don't have the capacity or time to fix It, then don't but for sure we need a separate issue for this function specifically!
150 lines of different edge cases is hard to follow.

We had two cases that were hard to follow - `target_roles` and `bump_all` and now we add two more...
Given that we have 4 arguments and each of them has its own separate `if` meaning they are not exclusive branches then questions like: what will happen if I call 
_`_update_snapshot(target_roles=["project-a", "project-b"], only_target=True, only_snapshot=True]`_
happen?

repository_service_tuf_worker/repository.py
```diff
@@ -308,29 +315,42 @@ def _bump_and_persist(
...
         self,
         target_roles: Optional[List[str]] = None,
         bump_all: Optional[bool] = False,
+        only_target: Optional[bool] = False,
+        only_snapshot: Optional[bool] = False,
```
MVrachev, line 353: You can simplify this function just by removing `only_snapshot`.

The previous implementation of `bump_snapshot(force=True)` did exactly what you want to achieve with this argument - it called `update_timestamp(update_snapshot())` without providing any argument.
This meant that `target_roles` or `bump_all` were not provided and it only bumped snapshot.

Maybe you can rename `bump_snapshot` to `bump_only_snapshot` or `bump_snapshot_without_targets()` or something like this and you will be able to remove `only_snapshot`.

MVrachev, line 353: I am not sure why you need to have both `only_target` and `only_snapshot`?
If `target` is not updated then you don't care even if you load it and do:
```
            snapshot.signed.meta[f"{Targets.type}.json"] = MetaFile(
                version=targets.signed.version
            )
            self._bump_and_persist(snapshot, Snapshot.type)
```
If `target` is changed wouldn't you want to always add the latest version to the new snapshot?

Wouldn't updating target and calling `update_snapshot(only_snapshot=True)` cause an inconsistent state as new snapshot will not contain the new target version in it's meta?

It's better always to load targets in snapshot meta just in case if there is a new version.

MVrachev (changes requested):
I have not yet finished, but this is a good start.
Will continue tomorrow.
I don't expect API and CLI to have so much to discuss.

repository_service_tuf_worker/repository.py
kairoaraujo: We have discussed refactoring this function for a while since you implemented the post-online rules bump ("bump_all").
While I was working on this PR, I started by refactoring this function, which resulted in a lot of changes in other parts that we don't want to make right now.
I think we can refactor it outside this PR. We can file an issue for it.

repository_service_tuf_worker/repository.py
kairoaraujo: I favor a full refactoring of this function but outside this PR.
I think we can add these comments in the same issue (with the https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/593#discussion_r1753162614)

kairoaraujo: It is true when update_snapshot was handling only artifacts, but now it handles Targets as a meta for Delegated roles

- When you are adding/removing/updating new Delegations and **all new delegations use offline keys**, you want only to update Target's metadata as the role that handles the delegation's information. (`targets_only`)
py#L881

 - When you are adding/removing/updating new Delegations and **delegations use offline keys and some delegations only online key**, you want only to update Target's metadata as the role that handles the delegation's information. (`targets_only` and `targets_roles`) 
 
 In both cases you cannot update it before calling `_update_timestamp(_update_snapshot...` because you don't know it was the state of the delegated roles
 
Again, a refactoring could help clarify it, but I prefer to do it in a separate and exclusive PR.

repository_service_tuf_worker/repository.py
MVrachev: Okay, as we are limited on time, lets do it in another pr after the conference.

repository_service_tuf_worker/repository.py
MVrachev: Okay, as we are limited on time, lets do it in another pr after the conference.

MVrachev: Okay, as we are limited on time, lets do it in another pr after the conference.

repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,388 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+
+            self._add_delegated_role_keys(
+                targets, delegations, role, role_metadata
+            )
+
+            if role not in targets.signed.delegations.roles:
+                logging.info(f"Role '{role}' added to Targets delegations")
+                targets.signed.delegations.roles[role] = delegations.roles[
+                    role
+                ]
```
MVrachev, line 871: If you call `add_metadata_delegations` then when will it happen that this `role` will be part from `targets.signed.delegations.roles`? 
Seems like this `if` will be always true.

repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,388 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+                self._update_snapshot(only_target=True), skip=True
+            )
+
+        return (success, failed)
+
+    def _update_metadata_delegation(
+        self,
+        delegations: Delegations,
+    ) -> Tuple[Dict[str, Metadata[Targets]], List[str]]:
+        """Create a custom delegation role"""
```
MVrachev, line 944:
```suggestion
        """Update a custom delegation role"""
```

repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,388 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+
+        targets = self._storage_backend.get(Targets.type)
+
+        if targets.signed.delegations.succinct_roles:
+            raise RepositoryError(
+                "Delegations already using hash-bins, cannot add custom roles"
```
MVrachev, line 950:
```suggestion
                "Delegations already using hash-bins, cannot update custom roles"
```

repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,388 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+            )
+
+        success = {}
+        failed = []
+        u_role: DelegatedRole  # updated role
+        o_role: DelegatedRole  # old role
+        logging.debug("adding roles to Targets delegations")
```
MVrachev, line 957:
```suggestion
        logging.debug("updating roles in Targets delegations")
```

repository_service_tuf_worker/repository.py
```diff
@@ -607,108 +679,388 @@ def save_settings(self, root: Metadata[Root], roles_info: Dict[str, Any]):
...
+        old_delegated: DelegatedRole,
+    ):
+        # NOTE: we don't support renaming and we verify it before
+        role_name = updated_delegated.name
+        # we can detect if keys are different, remove and add
+        if updated_delegated.keyids != old_delegated.keyids:
+            logging.debug("Identified different keys remove and add")
+
+            # remove keys
+            self._remove_delegated_role_keys(targets, old_delegated)
```
MVrachev, line 777: Why don't you add the new keys from `updated_delegated` after you have removed the old ones?

repository_service_tuf_worker/repository.py
```diff
@@ -1171,39 +1525,69 @@ def _run_online_roles_bump(self, force: Optional[bool] = False):
...
                 self._bump_and_persist(targets, Targets.type)
                 logging.info("Bumped version of 'Targets' role")
-                snapshot_bump = True
 
         if force:
             # Updating all delegated target roles.
-            timestamp = self._update_timestamp(
-                self._update_snapshot(bump_all=True)
+            self._update_timestamp(
+                self._update_snapshot(bump_all=True, only_target=True)
```
MVrachev, line 1541: Why we don't bump snapshot here?

repository_service_tuf_worker/repository.py
```diff
@@ -1171,39 +1525,69 @@ def _run_online_roles_bump(self, force: Optional[bool] = False):
...
+            else:
+                delegated_roles = list(targets.signed.delegations.roles.keys())
+
+            for role in delegated_roles:
+                if targets_crud.read_role_deactivated_by_rolename(
+                    self._db, role
+                ):
+                    delegated_roles.remove(role)
+                    logging.debug(f"Role '{role}' is deactivated, skipping")
+                    continue
```
MVrachev, line 1562: Maybe directly read from database the active roles?

repository_service_tuf_worker/repository.py
```diff
@@ -1680,69 +2161,102 @@ def _result(status, error=None, bootstrap=None, update=None):
...
-
         # Assert pending signing event exists
-        metadata_dict = self._settings.get_fresh("ROOT_SIGNING")
+        metadata_dict = self._settings.get_fresh(f"{rolename.upper()}_SIGNING")
         if metadata_dict is None:
-            msg = "No signatures pending for root"
+            msg = f"No signatures pending for {rolename}"
             return _result(False, error=msg)
 
         # Assert metadata type is root
```
MVrachev, line 2170:
```suggestion
```

MVrachev (changes requested):
I think those are my last set of requests for changes.
I finished my review.
I will not review the tests are they are an unfinished state.

Are we going to merge this with tests failing?

repository_service_tuf_worker/repository.py
kairoaraujo: When the user tries to add an existing role.

Also, if you see the `targets.signed.delegations.roles` is a list of DelegatedRoles, the user "could" add the same role twice.

repository_service_tuf_worker/repository.py
kairoaraujo: We do it using the `Delegation`

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/ee61ed4499d090933e77fdff4a7a631686925b28/repository_service_tuf_worker/repository.py#L976-L978

repository_service_tuf_worker/repository.py
kairoaraujo: We don't need it. When "bump_all" is used. If there is any change in any delegated role, it automatically bumps if there is any ch snapshot.

repository_service_tuf_worker/repository.py
kairoaraujo: We can discuss it.
I'm trying to avoid making the DB a source of trust. In that case, we use Metadata.
I think this is a good point for us to revisit.

repository_service_tuf_worker/repository.py
MVrachev: I wouldn't separate removing and adding keys in different functions.
Will add it to an issue.

MVrachev (dismissed):

repository_service_tuf_worker/repository.py
MVrachev: https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/595

kairoaraujo:
> Are we going to merge this with tests failing?

The only test failing is the coverage.

repository_service_tuf_worker/repository.py
MVrachev: https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/595

MVrachev (approved):

PR #600, opened 2024-09-14
build: allow to skip slow tests

kairoaraujo:
It gives us certain flexibility to skip slow when we identify that there is no changes that can affect the performance and skip the slow tests by adding the tag is a commit.

[skip slow]


MVrachev:
I am okay with having that option, but let's make sure always we skip slow tests to leave a comment in the pr about that and give a reason.

MVrachev (approved):

MVrachev:
Code coverage is not happy, but the coverage isn't lower now... I think it lowered based on your previous pr #593.

kairoaraujo:
> I am okay with having that option, but let's make sure always we skip slow tests to leave a comment in the pr about that and give a reason.

Yes, the good thing is that we keep the tests as required.
If something doesn't trigger it, we will see it. Like in this PR. ;)

kairoaraujo:
I will not merge it.
It is skipping all FT, where our intention is to skip only Slow DAS

Issue #598, opened 2024-09-13
Bug: MySQL is not working for custom delegation


kairoaraujo:
### What happened?

As required by MySQL, we need to review when we do `commit()` and make sure that the commits go in all async tasks.

See here a case:
https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/83d8937aa90e3b1a23bfdc6e468e1b9445ecf1e4/repository_service_tuf_worker/repository.py#L583-L588

The tests are failing

https://github.com/repository-service-tuf/repository-service-tuf-worker/actions/runs/10851053429/job/30113908168

### What steps did you take?

Run RSTUF Worker using MySQL

### Relevant log output

```shell
https://github.com/repository-service-tuf/repository-service-tuf-worker/actions/runs/10851053429/job/30113908168
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #601, opened 2024-09-14
fix: celery hostname for the instance

kairoaraujo:
It fixes the celery hotname instance.

We use this hostname to check the livenessprobe and readninessprobe in the helm charts.

https://github.com/repository-service-tuf/helm-charts/blob/176fa84eed2ed34a7ce22eabf7469425fdb42881/charts/rstuf-worker/values.yaml#L12

[skip slow]


MVrachev (approved):

kairoaraujo:
MySQL is failing as expected, see #598

PR #603, opened 2024-09-15
Remove deprecated and warn users for MySQL/MariaDB

kairoaraujo:
- remove deprecated code previously announced
- add note in the documentation for MySQL/MariaDB support
- disable tests for MySQL/MariaDB


MVrachev (dismissed):

kairoaraujo:
@MVrachev we need to reapprove this before #606

MVrachev (approved):

PR #629, opened 2024-10-18
Fix dependencies

kairoaraujo:
As the RSTUF worker is delivered as a container image, we don't require to
support multiple Python versions.

This commit removes multiple versions during the tests and uses 3.12

Aditional: update the latest dependencies


MVrachev (approved):

MVrachev (dismissed):

docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
         condition: service_healthy
 
   web:
-    image: python:3.10-slim-buster
+    image: python::3.12
```
MVrachev, line 51:
```suggestion
    image: python:3.12
```

MVrachev (approved):

PR #289, opened 2023-05-10
Improve RSTUF DB and operations.

kairoaraujo:
This commit improves the RSTUF SQL DB and uses the advantage of the database to make the TUF metadata operations smarter.

The changes:

* Create two tables `rstuf_target_roles` and rstuf_target_files` with a relationship between them.

  - It allows the possibility to be more flexible to the users by changing the number of delegated bins. RSTUF can update the DB and edit all target roles and files, persisting the new metadata in the backend storage
  - Update the CRUD to be more descriptive and use the relationship.
  - Makes more effective version handling for the target roles (BINS) with dynamic operations when persisting.

* Update the flow for `publish_targets` using the new DB design

  - Remove the operations of persisting the TUF metadata in the backend storage and add it to the `_update_snapshot`.
  - The `_update_snapshot` takes all the actions about persisting the TUF metadata for Target Role. It is more effective as the Snapshot updates the Meta with information about the Target Roles. `_update_snapshot` also updates the DB for the Target Role version and the Target Files published state.
  - The `_update_timestamp` is no longer responsible for updating the DB. It cares only about the Snapshot version.
  - Enables a new feature to the API to specify a list of Target Roles to be update

* Uses single lock `TARGETS_LOCK`

  - Remove the `TIMESTAMP_LOCK` usage, as every writing action is done by `_update_snapshot`. The repository uses the locks in three specific cases: `pulish_targets`, `bump_online_roles`, `metadata_rotation`

Note: During this, change has identified some refactorings, but it was limited to the database and publish target flows. It doesn't aim a refactoring in the repository.

Closes #276

PR #689, opened 2025-02-20
perf: fix bump online roles

kairoaraujo:
### The Problem
RSTUF right now has an issue if used the backend storage (TUF metadata) as S3 buckets with large number of delegations roles (example hash bins).

The problem starts with 512 delegation roles and above and it needs to update a large number delegation roles (bump online roles) at the same time.

An example, 2048 roles can tack ~60 min to update it. Usually it will fail due the timeout.

### Solution
Use Celery Chain to create multiple tasks that can be distributed to Workers|

In that PR we propose the use of Celery groups by using chunk feature and create a Chain.

Based in the above example it can bump 2048 roles in ~360 seconds (6 min) without hanging

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,20 +1,25 @@
 #!/bin/env python3
 
-# SPDX-FileCopyrightText: 2023 Repository Service for TUF Contributors
+# SPDX-FileCopyrightText: 2023-2025 Repository Service for TUF Contributors
 # SPDX-FileCopyrightText: 2022-2023 VMware Inc
 #
 # SPDX-License-Identifier: MIT
 
+import itertools
 import json
 import logging
+import time
 from enum import Enum
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 import redis
-from celery import Celery, schedules, signals
+from celery import Celery, chain, schedules, signals
 
 from repository_service_tuf_worker import get_worker_settings
-from repository_service_tuf_worker.repository import MetadataRepository
+from repository_service_tuf_worker.repository import (
+    MetadataRepository,
+    MetaFile,
+)
 
 logging.basicConfig(
     level=logging.DEBUG,
@@ -29,6 +34,9 @@ logging.basicConfig(
 
 
 worker_settings = get_worker_settings()
+
+BOR_LOCK = "BOR"
+BOR_TTL = worker_settings.get("BUMP_ONLINE_ROLES_TTL", 600)  # Lock expiration
 
 
 class status(Enum):
@@ -53,6 +61,9 @@ redis_backend = redis.StrictRedis.from_url(
 #     "ca_certs": "data/certs/ca_mq.crt",
 #     "cert_reqs": ssl.CERT_REQUIRED,
 # }
+
+repository = MetadataRepository.create_service()
+
 
 app = Celery(
     f"repository_service_tuf_worker_{worker_settings.WORKER_ID}",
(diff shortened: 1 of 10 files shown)


kairoaraujo:
Converting to Draft as we need to handle the custom delegation roles :(

app.py
```diff
@@ -71,6 +82,18 @@ class status(Enum):
 )
 
 
+def _calculate_chunk_size(num_roles: int) -> int:
+    chunk_size_cfg = repository._settings.get_fresh(
+        "BUMP_ONLINE_ROLES_CHUNK_SIZE", 500
+    )
+    if num_roles <= chunk_size_cfg:
+        chunk_size = int(num_roles / 2)
```
MVrachev, line 90: What if chunk_size_cfg is triple larger than num_roles?
For example: chunk_size = 50000, num_roles = 2000.
Should we have a more precise calculation here to make sure chunk_size will be indeed lower than num_roles?
Not a blocker, just a comment.

repository_service_tuf_worker/repository.py
```diff
@@ -295,19 +317,49 @@ def _bump_version(self, role: Metadata) -> None:
...
         """
-        self._bump_expiry(role, role_name)
+        self._bump_expiry(role, role_name, expire)
         self._bump_version(role)
-        self._sign(role)
+        self._sign(role, signer)
         if persist:
             self._persist(role, role_name)
 
+    def update_snapshot(
```
MVrachev, line 338: I am not a fan that we will have `_update_snapshot` which is not used in this function called  `update_snapshot`.

repository_service_tuf_worker/repository.py
```diff
@@ -1018,6 +1104,67 @@ def _update_metadata_delegation(
...
+        delegation.signed.targets.clear()
+        delegation.signed.targets = {
+            file.path: TargetFile.from_dict(file.info, file.path)
+            for file in db_role.target_files
+            if file.action == targets_schema.TargetAction.ADD
+            # Filtering the files with action 'ADD' cannot be done
+            # in CRUD. If a target role doesn't have any target
+            # files with an action 'ADD' (only 'REMOVE') then using
+            # CRUD will not return the target role and it won't be
+            # updated. An example can be when there is a role with
```
MVrachev, line 1117: But don't you do that here as well?
Now, when you filter by `if file.action == targets_schema.TargetAction.ADD`?

repository_service_tuf_worker/repository.py
```diff
@@ -1018,6 +1104,67 @@ def _update_metadata_delegation(
...
+        delegation: Metadata[Targets] = self._storage_backend.get(
+            rolename, db_target_role.version
+        )
+        self._update_delegated_role_target_files(delegation, db_target_role)
+        role_target_files = [file.path for file in db_target_role.target_files]
+
+        # Update expiry, bump version, and persist
+        self._bump_and_persist(
+            delegation, BINS, persist=False, signer=signer, expire=expire
+        )
```
MVrachev, line 1157: How do you know that the person is using `BINS` and not custom delegation?

MVrachev (commented):
Before approving can you please explain me some stuff from the pr that I added comments for.

I think really soon we will need to stop adding features and refactor `repository_service_tuf_worker/repository.py`

kairoaraujo:
> I think really soon we will need to stop adding features and refactor `repository_service_tuf_worker/repository.py`

I completely agree.
But, remember it is not a feature, this work here is a performance/bug fix.

I have already started some refactoring, but I avoided doing it together here.

Before lunching the 1.0.0, my goal is to do a good part of refactoring.
I'm first making sure we fix bug/performance issues with the existing code base and move the refactoring after that.

repository_service_tuf_worker/repository.py
kairoaraujo: This is not covered here.
I'll work to handle the case of custom delegations, not sure if already in this PR.

repository_service_tuf_worker/repository.py
kairoaraujo: First, this is a statement copy from here
https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/e5533ac15120a27cc4a93582930f04ddc7dd837f/repository_service_tuf_worker/repository.py#L457

What it means is that I could do CRUD filtering the db_roles object with action == 'ADD'
But there is a problem the DB query can return None the role as only 'REMOVED' and the delegated role (without any artifact ADD in action) will remain outdated.

We had discussed it here https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/289#discussion_r1198873147 (few years ago)

The goal idea is to refactor the `_update_snapshot`

repository_service_tuf_worker/repository.py
kairoaraujo: My goal is to refactor the `_update_snapshot`. I avoid doing it and right now we are adding the `update_snapshot` to avoid massive changes to it, that is already very complicated.

app.py
kairoaraujo: Ok, I will improve it.

kairoaraujo: Hi @MVrachev 
Yesterday evening I refactored it and added full UT for the tasks (forgot to push)
https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/689/commits/204158745e0058bc3dc37d06f5e42509bbcc9086

kairoaraujo:
Hi @MVrachev,

I will not push more changes to this PR.
I'm testing these changes in two "production" deployments, one with 4096 hash bins and another with custom delegations.

On top of this PR/branch, I will add PRs for refactoring `repository.py`.

Looking at this PR, besides the 1400 added lines, a big portion are tests and docs

repository_service_tuf_worker/repository.py
MVrachev: Okay, we need to get this done in the future.

Issue #708, opened 2025-03-23
RSTUF-CR-25-106: No Task Time Limit


kairoaraujo:
### What do you want to share with us?

No time limit is set on celery tasks. When these get stuck in an infinite loop or require overly
long to perform their task, this might lead to DoS situations.
Since no ways were identified to create a DoS situation, this is considered

### References

[Security Audit Report](https://github.com/user-attachments/files/19407542/X41-OSTIF-RSTUF-Audit-Final-Report-2025-02-14.pdf)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #743, opened 2025-05-15
added time limit to celery tasks

swastik959:
added timr limit to celery tasks in app.py for around an hour 
solves #708

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,10 @@ updates:
       prefix: "build"
       include: "scope"
     open-pull-requests-limit: 10
+    groups:
+      python-deps:
+        patterns:
+          - "*"    
   # Monitor Github Actions
   - package-ecosystem: "github-actions"
     directory: "/"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       run: tox
 
     - name: Codecov
-      uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d # v5.4.2
+      uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
       with:
         files: coverage.xml
         fail_ci_if_error: false
--- a/.github/workflows/publish_container.yml
+++ b/.github/workflows/publish_container.yml
@@ -48,7 +48,7 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Build and push
-      uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1
+      uses: docker/build-push-action@1dc73863535b631f98b2378be8619f83b136f4a0
       with:
         context: .
         push: true
--- a/.github/workflows/publish_docker_dev.yml
+++ b/.github/workflows/publish_docker_dev.yml
@@ -41,7 +41,7 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Build and push
-      uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1
+      uses: docker/build-push-action@1dc73863535b631f98b2378be8619f83b136f4a0
       with:
         context: .
         push: true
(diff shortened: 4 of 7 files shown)


app.py
```diff
@@ -77,6 +77,8 @@ class status(Enum):
     task_acks_late=True,
     task_track_started=True,
     broker_heartbeat=0,
+    task_time_limit=3600,
+    task_soft_time_limit=1800,
```
kairoaraujo, line 81: I was checking the Celery documentation https://docs.celeryq.dev/en/stable/userguide/workers.html#time-limits
Wouldn't be good to set it at the task level?

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/63c51c2813981277c600359c105714a9f6779906/app.py#L87

So we don't want the same limit for our internal tasks (bump online roles for example) and adding/removing artifacts for example.

Also, handle the soft time limit as described in the Celery documentation

I also would like to have it configurable by the user. Imagine that my use case I send 1000 new artifacts in one single request (which mean, one single task).

It can be easy done by[ documenting a env var RSTUF_<something>](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/63c51c2813981277c600359c105714a9f6779906/docs/source/guide/Docker_README.md) and using the <something> to se the parameter. There is few examples like 

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/63c51c2813981277c600359c105714a9f6779906/app.py#L39

Last, would be great adding the feature to our helm charts after merging it.

kairoaraujo (changes requested):
Hi @swastik959, 

Thank you for your contribution! 🎉 

I left a comment for us to elaborate more this PR

app.py
swastik959: @kairoaraujo added a few correction please review it now

swastik959:
@kairoaraujo I have done the relevant changes as suggested by you please review them

app.py
```diff
@@ -130,56 +138,68 @@ def _end_bor_chain_callback(result, start_time: float):
...
+    except SoftTimeLimitExceeded:
+        logging.warning(
+            f"Task _update_snapshot_timestamp reached soft time limit. "
+            "Cleaning up."
+        )
+        return {"status": "error", "message": "Task time limit exceeded."}
 
 
-@app.task(serializer="json", queue="rstuf_internals")
+@app.task(serializer="json", queue="rstuf_internals" , task_time_limit=TASK_TIME_LIMIT , task_soft_time_limit=TASK_SOFT_TIME_LIMIT)
```
kairoaraujo, line 202: I'm not sure if we should put the `task_[soft]_limit` for the `bump_online_role` right now, for some reasons:
- It will conflict directly with the `BOR_LOCK` we have to handle
- Depending on the number of delegated online roles, it will be a problem
- Using the same task limit time to the add/remove artifacts tasks

I'm in favor here of filing a new specific issue for adding task time limit for `bump_online_roles`

app.py
```diff
@@ -130,56 +138,68 @@ def _end_bor_chain_callback(result, start_time: float):
...
+    try:
+        return repository.update_targets_delegated_role(role)
+    except SoftTimeLimitExceeded:
+        logging.warning(
+            f"Task {role} reached soft time limit. Cleaning up."
+        )
+        return None
 
-@app.task(serializer="json", queue="rstuf_internals")
+@app.task(serializer="json", queue="rstuf_internals", task_time_limit=TASK_TIME_LIMIT , task_soft_time_limit=TASK_SOFT_TIME_LIMIT)
```
kairoaraujo, line 154: (same as `_update_online_role`)

app.py
```diff
@@ -130,56 +138,68 @@ def _end_bor_chain_callback(result, start_time: float):
     return {"result": result, "execution_time_seconds": total_time}
 
 
-@app.task(serializer="json", queue="rstuf_internals")
+@app.task(serializer="json", queue="rstuf_internals", task_time_limit=TASK_TIME_LIMIT , task_soft_time_limit=TASK_SOFT_TIME_LIMIT)
```
kairoaraujo, line 141: this task is a chain task from `bump_online_roles`.
I think we should use the same config here, see my comment on `bump_online_roles` -- I think it should part of a specific change from there.

kairoaraujo (changes requested):
I reviewed it and it looks good for `repository_service_tuf_worker` main tasks.
I just miss the UT for the `app.py` (see `tests/unit/test_app.py`)

Please, see my commends about the other tasks

Great work @swastik959

PR #753, opened 2025-05-22
fix: persist fail creating new custom delegation

kairoaraujo:
this bug happens when you have RSTUF deployed and bootstrapped with custom delegations

if you try to create new delegation the task fails while writing the metadata to the backend storage service.

It happens because persist is called without role name

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -309,9 +309,8 @@ class MetadataRepository:
         """
         Persists metadata using the configured storage backend.
 
-        The metadata role type is used as default role name. This is only
-        allowed for top-level roles. All names but 'timestamp' are prefixed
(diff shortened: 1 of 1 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -2209,8 +2221,8 @@ def metadata_delegation(
                 delegations: Delegations = Delegations.from_dict(
                     payload["delegations"]
                 )
-                targets = self._storage_backend.get(Targets.type)
-                snapshot = self._storage_backend.get(Snapshot.type)
+                targets = self._storage_load_targets()
```
kairoaraujo, line 2224: I will create a new good-first-issue to refactor the code using these help functions

lukpueh (approved):

Issue #789, opened 2025-08-10
Bug: Inconsistent Snapshot Metadata when adding or deleting custom delegations


srinjoydutta03:
### What happened?

In the `metadata_delegation` task, which is the function that handles the internal logic with regard to custom delegations updates, currently **does not** update the snapshot during metadata addition and deletion cases.

When a delegation is added or removed, the `snapshot` metadata object is correctly loaded and modified in memory within the `metadata_delegation` function's scope. However, before this modified object can be persisted, the helper function `_update_snapshot` is called. This helper function immediately reloads a fresh, unmodified copy of the snapshot from storage, effectively discarding all the in-memory changes.

As a result, the `snapshot.json` file that gets persisted has a bumped version and a new expiration date, but it contains the **old list of metadata files.**

This leads to two failure scenarios:

 - **On Add:** A new delegated role's metadata file (e.g., 1.project-a.json) is created, but the snapshot does not list it, making it invisible to TUF clients.
 - **On Delete**: A delegated role's entry is removed from the in-memory snapshot, but the persisted snapshot still contains the entry for the now-deleted role, leading to TUF client errors when they try to fetch a non-existent file or adding a new custom role with the same name later.

> It is to be noted that during bootstrap this feature correctly updates the snapshot, due the fact that the `_bootstrap_online_roles` function contains an online roles dictionary which contains the snapshot as an entry and later all the `online_roles` in that dictionary are individually persisted, thus giving us expected results.

### Relevant code snippets

[`metadata_delegation` function `add` case](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/e946e5f4cb1d7c4b83c9a7bc230e0bfa1139c8cf/repository_service_tuf_worker/repository.py#L2235-L2237)

[`_delete_metadata_delegation` function](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/e946e5f4cb1d7c4b83c9a7bc230e0bfa1139c8cf/repository_service_tuf_worker/repository.py#L1067-L1071)

## Scenario 1: Adding a custom delegation
- Perform the ceremony as usual
- Send a task to the `metadata_delegation` to create a new custom role named <_rolename_> through the API
- The task will report success
- Inspect the storage backend for the latest version of `snapshot.json`

## Scenario 2: Deleting a custom delegation
- Perform the ceremony as usual
- Send a task to the `metadata_delegation` to delete an existing custom role named <_rolename_> through the API
- The task will report success
[...]

PR #794, opened 2025-08-15
Fix snapshot state inconsistency after custom role add/delete

srinjoydutta03:
## Description
Added tests and persist calls in `_delete_metadata_delegation` and `metadata_delegation` function to properly persist snapshot and prevent inconsistent state when adding and deleting a custom role.

1) Updated existing tests for metadata deletion and addition
2) Added explicit test to check for correct snapshot persistence while adding
3) Placed `_persist` and `_bump_and_persist` calls at appropriate places to fix the issue

 - Minor Change: removed a redundant `_bump_and_persist` call in `_update_snapshot` function as already the flag `snapshot_meta_updated` is set in the `if` block for `only_target`, the same `_bump_and_persist` call happens when the control reaches the `if` block for `snapshot_meta_updated` in line number 579

Fixes: #789 

@kairoaraujo 
@MVrachev

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -151,7 +151,7 @@ class MetadataRepository:
         if key_dict is not None:
             return Key.from_dict(key_dict.pop("keyid"), key_dict)
         else:
-            root: Metadata[Root] = self._storage_backend.get(Root.type)
+            root: Metadata[Root] = self._storage_load_root()
             # All roles except root share the same one key and it doesn't
             # matter from which role we will get the key.
             keyid: str = root.signed.roles["timestamp"].keyids[0]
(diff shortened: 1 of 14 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -564,7 +564,6 @@ def _update_snapshot(
             snapshot.signed.meta[f"{Targets.type}.json"] = MetaFile(
                 version=targets.signed.version
             )
-            self._bump_and_persist(snapshot, Snapshot.type)
```
MVrachev, line 567: Removing this line here can lead to unexpected consequences as `_update_snapshot` with `only_target` flag can be called in multiple places.
Have you carefully researched when do we call `_update_snapshot` with `only_target`?

I didn't find another place besides `_add_metadata_delegation`, but still be really careful when changing a function that is used in multiple places such as `_update_snapshot`.

Have you tried instead of changing this here, to change https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/e946e5f4cb1d7c4b83c9a7bc230e0bfa1139c8cf/repository_service_tuf_worker/repository.py#L1025

to become: 

```
self._update_snapshot(target_roles=[success.keys()]), skip=True
```

The `success` dict contains names and Metadata values for new delegation roles.

PR #810, opened 2025-09-02
[Feature] Add support for hash-bin delegations under custom delegations

srinjoydutta03:
## Description

This pull request introduces initial support for nested hash-bin delegations under custom delegations, enhancing how custom delegation roles and their hash-bin children are created, managed, and deleted. The changes ensure that nested hash-bin roles are properly bootstrapped, persisted, and cleaned up, and that their metadata is correctly handled across all relevant operations.

**Support for nested hash-bin delegations:**

* Added `_setup_nested_hashbin_delegations` to create and persist nested hash-bin roles for custom delegations, including parallelized creation and database updates.
* Modified `_add_metadata_delegation` to detect custom roles requiring hash-bin delegation, invoke nested hash-bin setup, and return their metadata for further processing. 

**Handling and cleanup of nested roles:**

* Updated `_delete_metadata_delegation` to recursively deactivate and remove nested hash-bin child roles when their parent is deleted, ensuring proper cleanup.
* Enhanced `_run_online_roles_bump` to include nested hash-bin roles in expiry bump operations, maintaining their validity alongside parent roles.

**Integration with bootstrapping and delegation APIs:**

* Refined `_bootstrap_online_roles` and `metadata_delegation` to persist and register nested hash-bin roles in snapshots and metadata, ensuring they are available immediately after creation. 

**Key handling and snapshot updates:**

* Improved key lookup logic for hash-bin roles in `get_delegation_keyids` and `_update_snapshot`, ensuring correct key assignment for nested delegations. 
* Adjusted expiry bump logic to support hash-bin role naming conventions.

> Note: Currently this delegation pattern only supports the use of global online key for both the parent custom role and the delegated nested roles under it

**Role resolution improvements:**

* Enhanced `_get_role_for_artifact_path` to recursively resolve the most specific role for a given artifact, supporting the new delegation pattern

## Tests

All local tests are successfully passing as well as introduced unit tests to cover the feature. On the local development server the system behaves as expected, supporting all the intended operations and workflows.

Related Issue: https://github.com/repository-service-tuf/repository-service-tuf/issues/876

@kairoaraujo @MVrachev


MVrachev:
I will have time at the beginning of next week to review that.
Otherwise, awesome effort!

kairoaraujo (approved):

kairoaraujo (approved):

repository_service_tuf_worker/repository.py
```diff
@@ -932,12 +968,80 @@ def _update_delegated_roles(
...
+            for future in concurrent.futures.as_completed(futures):
+                rolename, metadata = future.result()
+                success[rolename] = metadata
+
+        targets_crud.create_roles(self._db, db_target_roles)
+        total_time = time.time() - start_time
+        logging.debug(
+            f"Added {len(db_target_roles)} nested hash bins \
+                      for role '{delegator_name} \
+                      in {total_time} seconds'"
```
MVrachev:
```suggestion
                      for role '{delegator_name}' \
                      in {total_time} seconds"
```

repository_service_tuf_worker/repository.py
```diff
@@ -932,12 +968,80 @@ def _update_delegated_roles(
...
+
+            db_target_roles.append(
+                targets_schema.RSTUFTargetRoleCreate(
+                    rolename=bin_rolename,
+                    version=1,
+                    expires=bin_metadata.signed.expires,
+                )
+            )
+            self._persist(bin_metadata, bin_rolename)
+            return bin_rolename, bin_metadata
```
MVrachev: Instead of appending to a shared `db_target_roles` list which is executed in multiple threads, it's better to ensure thread-safety and that no threads will write at the same container at once.

This can be achieved if you return this new `db_target_role` result from `process_delegated_bin_roles()` and separately call the thread pool completed like this:

```
       def process_delegated_bin_roles(bin_rolename: str):
            bin_metadata = Metadata(Targets(version=1))
            self._bump_expiry(bin_metadata, name_prefix, expire=expire_bins)
            self._sign(bin_metadata, signer)
            db_role = targets_schema.RSTUFTargetRoleCreate(
                rolename=bin_rolename,
                version=1,
                expires=bin_metadata.signed.expires,
            )
            self._persist(bin_metadata, bin_rolename)
            return bin_rolename, bin_metadata, db_role

        db_target_roles: List[targets_schema.RSTUFTargetRoleCreate] = []
        with concurrent.futures.ThreadPoolExecutor() as executor:
            start_time = time.time()
            futures = [
                executor.submit(process_delegated_bin_roles, bin_rolename)
                for bin_rolename in succinct_roles.get_roles()
            ]
            for future in concurrent.futures.as_completed(futures):
                rolename, metadata, db_role = future.result()
                success[rolename] = metadata
                db_target_roles.append(db_role)

```

We need to be careful when working with threads.

MVrachev (changes requested):
Great job overall, but there are a few changes needed.

repository_service_tuf_worker/repository.py
```diff
@@ -932,12 +968,80 @@ def _update_delegated_roles(
                 f"{role_name.upper()}_EXPIRATION", expires
             )
 
+    def _setup_nested_hashbin_delegations(
+        self,
+        delegator_metadata: Metadata[Targets],
+        delegator_name: str,
+        expiration: int,
+        num_bins: int,
+    ):
```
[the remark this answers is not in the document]
srinjoydutta03: Added return type

repository_service_tuf_worker/repository.py
```diff
@@ -932,12 +968,80 @@ def _update_delegated_roles(
...
+        bit_length = int(log(num_bins, 2))
```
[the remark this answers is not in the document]
srinjoydutta03: Added check for the same

repository_service_tuf_worker/repository.py
```diff
@@ -220,11 +220,22 @@ def get_delegation_keyids(self, rolename: str) -> List[str]:
             # we know all roles uses the online key
             delegation_keyids = [self._online_key.keyid]
         else:
-            logging.debug("delegations using custom delegations")
-            targets: Metadata[Targets] = self._storage_load_targets()
-            delegation_keyids = targets.signed.delegations.roles[
-                rolename
-            ].keyids
+            logging.debug(f"delegations using custom delegations {rolename}")
+            if "-bins-" in rolename:
```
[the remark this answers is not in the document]
srinjoydutta03: Declared constant at the top of the file

repository_service_tuf_worker/repository.py
srinjoydutta03: Refactored

repository_service_tuf_worker/repository.py
```diff
@@ -986,30 +1091,79 @@ def _add_metadata_delegation(
...
+                        )
+                    ):
+                        logging.debug(
+                            f"Role '{role}' has nested \
+                                      hash-bin delegations"
+                        )
+                        delegations.roles[role].terminating = False
+                        nb_success = self._setup_nested_hashbin_delegations(
+                            role_metadata, role, expires, num_bins
+                        )
```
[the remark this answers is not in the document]
srinjoydutta03: Updated

repository_service_tuf_worker/repository.py
srinjoydutta03: Updated

repository_service_tuf_worker/repository.py
```diff
@@ -1245,17 +1265,18 @@ def _delete_metadata_delegation(
...
 
         if success:
             self._bump_and_persist(targets, Targets.type)
-            self._update_timestamp(
-                self._update_snapshot(only_target=True), skip=True
+            snapshot.signed.meta[f"{Targets.type}.json"] = MetaFile(
+                version=targets.signed.version
             )
+            self._bump_and_persist(snapshot, Snapshot.type)
+            self._update_timestamp(snapshot.signed.version)
```
srinjoydutta03, line 1301: @MVrachev Added a little optimization here, instead of N DB calls for N roles, i update the snapshot and persist it at the end. Since,  for large number of bins it will be performance intensive.

repository_service_tuf_worker/repository.py
```diff
@@ -2420,6 +2441,10 @@ def metadata_delegation(
                 # helper function `_add_metadata_delegation` because while
                 # running `_bootstrap_online_roles` targets still not exist
                 # and require to be persisted before
+                if len(success) > 0:
+                    snapshot.signed.meta[f"{Targets.type}.json"] = MetaFile(
+                        version=targets.signed.version
+                    )
```
srinjoydutta03, line 2468: The reason for this, is when testing i observed that for a single metadata delegation add, two snapshot metadata files was persisted. One containing the latest version for targets after it was updated and the next one included all of the metadata files in it. So, the latest one is correct

This fix, makes sure that all of that happens in a single snapshot update (Tested and verified locally)

repository_service_tuf_worker/repository.py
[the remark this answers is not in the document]
MVrachev: If this were CLI, it would be good to fail loudly when `num_bins` is not a power of 2, but for the worker, this solution is fine.

repository_service_tuf_worker/repository.py
MVrachev: That optimization sounds reasonable, and you can use it to verify all bins were updated successfully.
Now, you update the snapshot if `success` is non-empty, but do you want to update it when ALL bins were successfully updated instead?
But think carefully about how to make this check.

repository_service_tuf_worker/repository.py
MVrachev: Did you add a specific test for that?

MVrachev (approved):

PR #843, opened 2026-03-31
fix: reject delegation roles with overlapping paths

### LINKED ISSUE #782: Bug: Custom Delegated Roles with overlapping paths cause incorrect artifact role resolution while addition
### What happened?

When you add a new custom delegation whose path patterns overlap those of an existing role, the service accepts it without complaint. For example, if role A has paths `["project-a/"]` and you later add role B with `["project-a/"]`, both roles coexist. 

This leads to ambiguous matching, so artifacts may be routed to the wrong role (or may cause inconsistencies later). Specifically, the artifact gets routed to the first delegated role with corresponding path.

Updated `project-a.json` metadata
*Figure 1: Updated `project-a.json` metadata showing the delegated role structure after artifact addition.*

Targets metadata file depicting two custom roles with overlapping paths
*Figure 2: `targets.json` showing two roles with overlapping path patterns.*
Image

### What steps did you take?

_No response_

The service should detect any prefix-overlap between the new delegation’s paths and existing custom-delegation paths (or among new paths themselves).

If any overlap is found, the `add_delegation` operation must fail immediately, and the client should receive a failure response instead of proceeding with an inconsistent state.

### Relevant log output

```shell

```

### Code of Conduct

Sandipmandal25:
### Description

`_add_metadata_delegation` validated role name uniqueness but did not check whether incoming role paths overlap with those already claimed by existing delegated roles.

Since TUF resolves delegations using first-match ordering, overlapping path patterns can silently coexist. In such cases, artifacts are always routed to the first matching role, making subsequent roles effectively unreachable.

This PR introduces a path-overlap validation step before accepting new delegated roles:

1. Collect all paths from existing delegated roles before processing new entries  
2. Reject any incoming role whose paths intersect with the existing set  
3. Update the path set after each successful addition to also detect intra-batch overlaps within the same request  

Rejected roles are appended to the failed list with a descriptive reason, consistent with the handling of duplicate role names.

### Tests

- `test__add_metadata_delegation_rejects_overlapping_paths`  
  Ensures a role with a path already claimed by an existing role is rejected  

- `test__add_metadata_delegation_accepts_non_overlapping_paths`  
  Ensures a role with a unique path is accepted  

- `test__add_metadata_delegation_rejects_intra_batch_overlapping_paths`  
  Ensures that when two roles in the same request share a path, the first is accepted and the second is rejected  
image

All 234 unit tests pass locally.

image

### Fixes

Fixes: #782

### Reviewers

@srinjoydutta03 
@kairoaraujo   
@MVrachev

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -952,6 +952,12 @@ class MetadataRepository:
         failed = []
         db_roles = []
 
+        existing_paths: set = set()
+        if targets.signed.delegations.roles:
+            for existing_role in targets.signed.delegations.roles.values():
+                if existing_role.paths:
+                    existing_paths.update(existing_role.paths)
+
         logging.debug("adding roles to Targets delegations")
         for role in delegations.roles:
             if targets.signed.delegations.roles is None:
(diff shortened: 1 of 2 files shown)


OmBiradar:
Hey @Sandipmandal25 !

Nice work there. 

One suggestion I have is - have you confirmed this is consistent with TUF (I mean The Update Framework) specs?

https://github.com/theupdateframework/specification

It would be great to know if it matches the spec.

PR #845, opened 2026-04-01
fix: copy online key dict before mutating in _online_key property

### LINKED ISSUE #431: Bug: public-to-private key mapping is not reliable in LocalKeyVault
### What happened?

`LocalKeyVault.get` returns a "matching" private key for a passed public key. Keys are assumed to match, if the private key object can be constructed with the public key object as input:

https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/a3f4553bee8e588823b51328d9a6b454746397b7/repository_service_tuf_worker/services/keyvault/local.py#L198-L207

This assumption is invalid, because the constructor factory does not guarantee to fail, if keys don't match. 

As a consequence, the produced signatures may become unverifiable. This is because signatures are assigned the keyid of the public key attached to the used signer. This keyid is then used at verification time by the TUF client to find the matching public key in TUF delegator metadata. If the public key with that keyid didn't match at signing time it won't match at verification time either.

### What steps did you take?

_No response_

Reliably map public and private keys using public key keyids in private key configuration.

### Relevant log output

_No response_

### Code of Conduct

Sandipmandal25:
## Description

This PR fixes two issues in the worker's online key handling, both part of the remaining gaps tracked in #431.

### 1. `_online_key` mutates shared settings dict (repository.py)

`Key.from_dict(key_dict.pop("keyid"), key_dict)` was calling `.pop()` directly on the dict returned from settings.  
This mutated the shared object in place, so subsequent calls to `_online_key` received a dict with `keyid` already removed, leading to inconsistent state.

**Fix:**  
Create a copy of the dict before calling `.pop()` to avoid mutating shared state.

### 2. `SignerStore.get()` raises unclear `KeyError` when URI is missing (signer.py)

If the `x-rstuf-online-key-uri` field was missing from the online key metadata (e.g. due to a ceremony run with an older CLI), `SignerStore.get()` skipped loading the signer and later raised a `KeyError` during lookup.

This made the failure mode hard to diagnose.

**Fix:**  
Add an explicit check for the missing field and raise a `ValueError` with a clear message indicating that the issue likely originates from the ceremony step.

Fixes #431

## Type of change

- Bug fix (non-breaking change)

## Additional requirements

- [x] Tests have been added for the bug fix

## Code of Conduct

**note** 
Worker pr  [#845 ](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/845) fixes the dict mutation `(.pop("keyid")` modifying shared state) and adds a clear error message when `x-rstuf-online-key-uri `is missing
CLI pr [#918](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/918) — fixes the ceremony writing the key identifier instead of `fn:<filename> `into `x-rstuf-online-key-uri`, so the worker can actually find the key file at signing time

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -149,7 +149,7 @@ class MetadataRepository:
     def _online_key(self) -> Key:
         key_dict = self._settings.get_fresh("ONLINE_KEY")
         if key_dict is not None:
-            key_dict = key_dict.copy()
+            key_dict = copy.deepcopy(key_dict)
             return Key.from_dict(key_dict.pop("keyid"), key_dict)
         else:
             root: Metadata[Root] = self._storage_load_root()
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -149,6 +149,7 @@ class MetadataRepository:
     def _online_key(self) -> Key:
         key_dict = self._settings.get_fresh("ONLINE_KEY")
         if key_dict is not None:
+            key_dict = key_dict.copy()
             return Key.from_dict(key_dict.pop("keyid"), key_dict)
         else:
             root: Metadata[Root] = self._storage_load_root()
--- a/repository_service_tuf_worker/signer.py
+++ b/repository_service_tuf_worker/signer.py
@@ -125,11 +125,17 @@ class SignerStore:
         """
 
         if key.keyid not in self._signers:
-            if uri := key.unrecognized_fields.get(RSTUF_ONLINE_KEY_URI_FIELD):
-                # (Re-)export ambient settings in isolated environment
-                with isolated_env(self._ambient_settings):
-                    signer = Signer.from_priv_key_uri(uri, key)
+            uri = key.unrecognized_fields.get(RSTUF_ONLINE_KEY_URI_FIELD)
+            if not uri:
+                raise ValueError(
+                    f"Online key {key.keyid!r} is missing the "
+                    f"'{RSTUF_ONLINE_KEY_URI_FIELD}' field. "
+                    "Re-run the TUF ceremony to set the correct key URI."
+                )
+            # (Re-)export ambient settings in isolated environment
+            with isolated_env(self._ambient_settings):
+                signer = Signer.from_priv_key_uri(uri, key)
 
-                self._signers[key.keyid] = signer
+            self._signers[key.keyid] = signer
 
         return self._signers[key.keyid]
--- a/tests/unit/tuf_repository_service_worker/test_repository.py
+++ b/tests/unit/tuf_repository_service_worker/test_repository.py
@@ -102,7 +102,7 @@ class TestMetadataRepository:
     def test_online_key_property_does_not_mutate_settings_dict(
         self, test_repo, monkeypatch
     ):
-        fake_key_dict = {"keyval": "foo", "keyid": "key_id"}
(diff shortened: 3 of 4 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -149,6 +149,7 @@ def _settings(self) -> Dynaconf:
     def _online_key(self) -> Key:
         key_dict = self._settings.get_fresh("ONLINE_KEY")
         if key_dict is not None:
+            key_dict = key_dict.copy()
```
MVrachev: Better do a deepcopy shallow copy could be a problem as from securesystemslib a key can contain:

```
{
    "keyid": "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
    "keytype": "ed25519",
    "scheme": "ed25519",
    "keyval": {
        "public": "a1b2c3d4e5f6..."
    },
    "x-rstuf-online-key-uri": "fn:ed25519_private.pem"
}
```

And because of `keyval` value, you have a nested dict which will not be copied, but it will be a shared reference between the original dict and the copy.

MVrachev (commented):
Thanks for the pr!

Still, I don't see you actually referencing the main issue - the #431, which you are referencing.

Sandipmandal25:
> Thanks for the pr!
> 
> Still, I don't see you actually referencing the main issue - the #431, which you are referencing.

Thanks for the review!

created a gist for better understanding -https://gist.github.com/Sandipmandal25/62f82cf60385ced0a5cd4a7d99851c47

Just to give some context on the #431 reference so this pr fix the remaining gaps so while exploring the signing flow I traced the design across Worker issue #431 CLI issue [#642](https://github.com/repository-service-tuf/repository-service-tuf-cli/issues/642) and Umbrella issue [#580](https://github.com/repository-service-tuf/repository-service-tuf/issues/580). The discussion in that issue[#580](https://github.com/repository-service-tuf/repository-service-tuf/issues/580) converged on Option 2 that encoding the key location in `x-rstuf-online-key-uri` as `fn:<filename>`. and the same i have discussed with justin and said to follow up in slack.

The worker side already supports reading this URI but the CLI ceremony was writing the key identifier instead of the filename, so the worker could not locate the key at signing time. and this pr addresses the dict mutation side effect and adds a clearer error when the URI is missing. and added pr in CLI side  [#918](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/918) that fixes the ceremony to write the correct value. Together, they align with the Option 2 approach from discussion on [#580](https://github.com/repository-service-tuf/repository-service-tuf/issues/580) .

I also noticed @lukpueh  draft pr [#608](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/608) and left a comment there too about the same.

repository_service_tuf_worker/repository.py
Sandipmandal25: yes keyval is nested so a shallow copy leaves it as a shared reference. Will update this!

Sandipmandal25:
also @MVrachev the original `LocalKeyVault.get()` issue in the issue #431 was resolved in pr  #507 where `LocalKeyVault` was removed. with the commit sms
image


and my 2 prs fix the remaining gaps what i discussed in meet the `_online_key` dict mutation and incorrect URI from the ceremony so the `SignerStore` + `FileNameSigner` flow works end to end.


Also @OmBiradar if you leave your thoughts in this because this is what i mentioned in the meet!

Sandipmandal25:
covered the complete thing here - https://gist.github.com/Sandipmandal25/62f82cf60385ced0a5cd4a7d99851c47

PR #853, opened 2026-04-05
feat: resolve delegated-role signers via URI or global online key

geeky33:
### **PR Description**
**Summary**
Introduce delegation-aware signer resolution for delegated targets. Roles now prefer their own `x-rstuf-online-key-uri` keys for automated signing, falling back to the repository global online key when present. This removes the implicit dependency on a single global key.

**Changes**

* Added `get_signer_for_delegation` and `delegation_uses_automated_signing`
* Used `succinct_roles.keyids` for hash-bin delegations
* Passed resolved signer through `_update_snapshot`, `bump_persist_role`, and `_add_delegated_role_keys`
* Allowed optional `signer` in `_sign` (default unchanged for top-level roles)
* Fixed `_online_key` to avoid mutating cached settings

**Tests**

* Cover signer resolution (role URI vs global fallback)
* Validate succinct keyid handling
* Update `bump_persist_role` to assert signer wiring

image


Sandipmandal25:
Hey @geeky33 just a heads up the `_online_key` dict mutation fix in this PR duplicates my PR #845 which was opened earlier and addresses the same bug. Could you rebase on top of #845 or remove that part to avoid the overlap?
also please follow the new pr desc template. Thank you!

repository_service_tuf_worker/repository.py
```diff
@@ -1166,29 +1204,28 @@ def bump_persist_role(
         from_storage: bool,
     ):
 
+        targets: Metadata[Targets] = self._storage_load_targets()
         delegation_keyids = self.get_delegation_keyids(rolename)
+        delegation_signer = self.get_signer_for_delegation(
+            targets, delegation_keyids
+        )
```
Sandipmandal25, line 1210: Noticed that `bump_persist_role` loads `targets` at the top but `get_delegation_keyids()` for custom delegations also calls `_storage_load_targets()` internally so for every custom delegation bump the storage backend is hit twice and the second load is effectively discarded; would it make sense to pass the already loaded `targets` into `get_delegation_keyids` to avoid the redundant I/O? What do you think @geeky33 @kairoaraujo ?

repository_service_tuf_worker/repository.py
```diff
@@ -215,19 +219,49 @@ def _storage_load_root(self) -> Metadata[Root]:
...
+        targets: Metadata[Targets],
+        delegation_keyids: List[str],
+    ) -> Optional[Signer]:
+        """Resolve the signer used for automated (online) delegation signing.
+
+        Prefer a role key that carries ``x-rstuf-online-key-uri``; otherwise
+        use the repository global online key if it is listed for the role.
+        """
+        keys_map = targets.signed.delegations.keys
+        for keyid in delegation_keyids:
```
Sandipmandal25, line 246: `targets.signed.delegations` is optional per the TUF spec so it can be `None` and `targets.signed.delegations.keys` would raise an `AttributeError` in that case. i think tho a small guard like `if targets.signed.delegations is None: return None` at the top of the method would make this safe.

repository_service_tuf_worker/repository.py
```diff
@@ -215,19 +219,49 @@ def _storage_load_root(self) -> Metadata[Root]:
...
+            if keyid == self._online_key.keyid:
+                return self._signer_store.get(self._online_key)
+        return None
+
+    def delegation_uses_automated_signing(
+        self,
+        targets: Metadata[Targets],
+        delegation_keyids: List[str],
+    ) -> bool:
+        return self.get_signer_for_delegation(targets, delegation_keyids) is not None
```
Sandipmandal25, line 262: umm this method is only called in tests and not in any production code path...
callers could just do the `is not None` check directly on `get_signer_for_delegation`
rather than adding a wrapper to the class, unless theres a plan to use it somewhere
down the line? curious what the intent is here @geeky33

tox.ini
```diff
@@ -32,8 +32,13 @@ commands =
...
 
 [testenv:test]
+# -n auto uses one worker per CPU and can exhaust RAM on high-core laptops
+# (looks like a full system hang). Default caps workers; override when needed:
+#   tox run -e test -- -n auto
+# Serial (no xdist), e.g. if it still freezes:
+#   tox run -e test -- -n 0
 commands =
-    python -m pytest --cov-report=xml --cov-report=term --cov-config=tox.ini --cov -n auto -vv tests/
+    python -m pytest --cov-report=xml --cov-report=term --cov-config=tox.ini --cov -n 2 -vv tests/ {posargs}
```
Sandipmandal25, line 41: the  `-n auto` to `-n 2` changes are unrelated to the delegation signer feature... the `-n 2` change especially could slow down CI on runners with more cores... context for adding these changes @geeky33 ??

Pipfile
```diff
@@ -11,7 +11,7 @@ dynaconf = {extras = ["ini"], version = "*"}
 supervisor = "*"
 securesystemslib = "*"
 sqlalchemy = "*"
-psycopg2 = "*"
+psycopg2-binary = "*"
```
Sandipmandal25, line 14: unrelated chnages i guess!!

Pipfile.lock
```diff
@@ -928,15 +987,75 @@
...
+                "sha256:31b32c457a6025e74d233957cc9736742ac5a6cb196c6b68499f6bb51390bd6a",
+                "sha256:32770a4d666fbdafab017086655bcddab791d7cb260a16679cc5a7338b64343b",
+                "sha256:366df99e710a2acd90efed3764bb1e28df6c675d33a7fb40df9b7281694432ee",
+                "sha256:37d8412565a7267f7d79e29ab66876e55cb5e8e7b3bbf94f8206f6795f8f7e7e",
+                "sha256:4012c9c954dfaccd28f94e84ab9f94e12df76b4afb22331b1f0d3154893a6316",
+                "sha256:41360b01c140c2a03d346cec3280cf8a71aa07d94f3b1509fa0161c366af66b4",
+                "sha256:44fc5c2b8fa871ce7f0023f619f1349a0aa03a0857f2c96fbc01c657dcbbdb49",
+                "sha256:47f212c1d3be608a12937cc131bd85502954398aaa1320cb4c14421a0ffccf4c",
+                "sha256:4bdab48575b6f870f465b397c38f1b415520e9879fdf10a53ee4f49dcbdf8a21",
+                "sha256:4dca1f356a67ecb68c81a7bc7809f1569ad9e152ce7fd02c2f2036862ca9f66b",
```
Sandipmandal25, line 1010: pipfile.lock? I think it should not be committed here unless updating dependencies maybe you updated pipfile psycopg2-binary and ran pipenv install locally and pipfile.lock generated and commited btw It looks unrelated to the feature and adds a lot of lockfile noise.

Sandipmandal25 (changes requested):
signer changes looks good but Pipfile / Pipfile.lock and tox.ini changes seem unrelated to this PR can you clean up this pr so this one stays focused? and please remove the redundancy with #845

PR #862, opened 2026-04-08
Add mypy checks and type annotations                

### LINKED ISSUE #365: Task: introduce type annotations and mypy checks
### What is the task about?

Similar to changes proposed by @KAUTH in https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/181 we can introduce annotations to improve code quality, faster bug discovery and features like in #362.

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

theycallmeaabie:
Closes #365                                                                                                                           
                                                                                                                                      
Adds mypy configuration, integrates it into the CI pipeline, fixes existing type errors, and adds return type annotations across the codebase.                                                                                                                             
 
### Changes                                                                                                                           
- Added `[mypy]` config section in `tox.ini` (Python 3.13, ignore missing imports)
- Added mypy pre-commit hook via `mirrors-mypy`                                                                                       
- Added mypy to `[testenv:lint]` commands
- Excluded `alembic/` and `docs/` directories from mypy checks                                                                        
- Fixed 35 existing type errors in `repository.py` and `app.py`
- Added return type annotations to all unannotated functions


vickysharma-prog:
Thanks for working on this! Adding mypy is a valuable step for the project.

I went through the diff carefully against the actual codebase and noticed a few things that might be worth revisiting:

**1**. get_delegated_rolenames return type:**The change to `List[Optional[str]]` seems inaccurate - looking at the actual function body, it always appends string values (`Targets.type` and DB role names), so None is never actually returned. The original `List[str]` appears correct here. The `if r is not None` guard in app.py looks like a defensive check rather than something reflecting the actual return type.

**2**. _update_snapshot_timestamp return type:**The change to `-> None` might be misleading since this function is used as part of a Celery chain:
chain(
    task_groups,
    _update_snapshot_timestamp.s(...),
    _end_bor_chain_callback.s(start_time),
)
Celery tasks implicitly return results through the chain - `Dict[str, Any]` seemed more representative of the actual behavior.

**3**. ignore_missing_imports = True (global):**This works but hides missing stubs for all packages globally. A more precise 
approach could be per-module ignores for packages that genuinely lack stubs:

[mypy-dynaconf.*]
ignore_missing_imports = True
[mypy-tuf.*]
ignore_missing_imports = True
[mypy-securesystemslib.*]
ignore_missing_imports = True

Also `types-redis` and `celery-types` are available on PyPI and could be added to `additional_dependencies` in the mypy 
pre-commit hook for more accurate checking.

Happy to help if any of these need further discussion!

MVrachev:
I will address what @vickysharma-prog is saying and you can address the required changes @theycallmeaabie.

> 1. get_delegated_rolenames return type:**The change to List[Optional[str]] seems inaccurate - looking at the actual function body, it always appends string values (Targets.type and DB role names), so None is never actually returned. The original List[str] appears correct here. The if r is not None guard in app.py looks like a defensive check rather than something reflecting the actual return type.

He is right, if you look closer at the `targets_crud.read_roles_rolenames_expired` implemention from where you can theoretically you can get `None` you will see that actually the None case is caught. So, make the `get_delegated_rolenames` return annotation to just `List[str]` and fix `read_roles_rolenames_expired` return annotation to `List[str]`  as well.

> 2. _update_snapshot_timestamp return type:**The change to -> None might be misleading since this function is used as part of a Celery chain:
chain(
task_groups,
_update_snapshot_timestamp.s(...),
_end_bor_chain_callback.s(start_time),
)
Celery tasks implicitly return results through the chain - Dict[str, Any] seemed more representative of the actual behavior.

I don't think that is correct, the `None` annotation seems correct as there is no `return` or `yeild` instruction.

> 3. ignore_missing_imports = True (global):**This works but hides missing stubs for all packages globally. A more precise
approach could be per-module ignores for packages that genuinely lack stubs:

> [mypy-dynaconf.]
ignore_missing_imports = True
[mypy-tuf.]
ignore_missing_imports = True
[mypy-securesystemslib.*]
ignore_missing_imports = True

Yes, I will agree it's better to list them.
If they are many let us know.

> Also types-redis and celery-types are available on PyPI and could be added to additional_dependencies in the mypy
pre-commit hook for more accurate checking.

When adding dependecies, please be careful if they come from trusted developerrs. For `typos-redis` and `celery-types` you need to check that the developers are also behind redis and Celery respectively.

vickysharma-prog:
Thanks @MVrachev!

On the `_update_snapshot_timestamp` point you're absolutely right and I appreciate the correction. I got confused between 
how Celery passes results through a chain internally versus what the function actually returns at the Python level. Since 
there's no explicit return statement in the function body, `-> None` is the correct annotation. That was a misunderstood by me.

Regarding `types-redis` and `celery-types` I think @theycallmeaabie should verify the publishers directly on 
PyPI before adding them as dependencies, especially keeping your concern in mind about whether they come from the official maintainers of redis and Celery respectively. 

If any help is needed at any point, happy to help @theycallmeaabie!

theycallmeaabie:
Thanks @vickysharma-prog and @MVrachev, sorry for the delay on this. Just pushed a commit addressing the feedback

MVrachev:
Seems that `lint` and `test` fail.
Can you research that?

theycallmeaabie:
Done, it should pass now

MVrachev:
@theycallmeaabie, please fix the linting and run it before pushing.

PR #880, opened 2026-04-14
feat: implement FT tests when a PR is approved

### LINKED ISSUE #258: Task: Implement FT tests when a PR is approved
### What is the task about?

This issue aims to implement functional test (FT) checks when a PR has a review approved.

When a maintainer does the first approval for the PR, it will trigger the FT. I should use the "dev" version for all components.

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

yashasviyadav30:
## What is the PR about?
This PR implements functional test (FT) checks when a PR has a review approved, as described in issue #258.

When a maintainer approves a PR, the review-approved.yml workflow triggers FT tests using the dev version of other components (API, CLI) while using the local build context for the worker to ensure the PR changes are correctly validated.

**Changes:**
- Add `worker_version` input to `functional-tests.yml` workflow (both `workflow_dispatch` and `workflow_call`)
- Pass `WORKER_VERSION` env variable to the compose action in all jobs
- Update `review-approved.yml` to pass `worker_version: dev` on PR approval
- Ensure local build context (`build: context: .`) and volumes are maintained in docker-compose files to test PR-specific changes
- Refactor `docker-compose.yml` DB connection string for consistency with other compose files

## Fixes
Closes #258

## Types of changes
* [x] New feature (non-breaking change which adds functionality)

## Code of Conduct
* [x] I agree to follow this project's Code of Conduct


docker-compose.yml
```diff
@@ -63,21 +63,15 @@ services:
       - "6379:6379"
 
   repository-service-tuf-worker:
-    build:
-      context: .
-    entrypoint: "bash entrypoint-dev.sh"
+    image: ghcr.io/repository-service-tuf/repository-service-tuf-worker:${WORKER_VERSION}
```
Sandipmandal25: `${WORKER_VERSION}` has no default so `make run-dev` / `make ft-das` will fail locally since the Makefile never sets this. can use `${WORKER_VERSION:-dev}` instead ? cc @kairoaraujo

Sandipmandal25 (commented):
just one thing on the docker-compose files need to look.

docker-compose.yml
yashasviyadav30: Thanks! Fixed with WORKER_VERSION defaults

yashasviyadav30:
@MVrachev @kairoaraujo PR is ready for review whenever you get a chance!

docker-compose.yml
```diff
@@ -63,21 +63,15 @@ services:
       - "6379:6379"
 
   repository-service-tuf-worker:
-    build:
-      context: .
-    entrypoint: "bash entrypoint-dev.sh"
+    image: ghcr.io/repository-service-tuf/repository-service-tuf-worker:${WORKER_VERSION:-dev}
```
MVrachev: I believe this should stay; Docker Compose is used for building a container image locally as well.

Why do you decide to pull the latest pushed dev tag:
ghcr.io/repository-service-tuf/repository-service-tuf-worker:${WORKER_VERSION:-dev}
and not build it with the latest pr changes?

I think it should instead use the current context (as it's before the change) and use changes from the pr to run the ft tests.

Makefile
```diff
@@ -7,6 +7,7 @@ DC := docker-compose.yml
 endif
 
 API_VERSION ?= dev
+WORKER_VERSION ?= dev
```
MVrachev: Makefile is used for local development and I am not sure what this helps here.

docker-compose.yml
Sandipmandal25: yes @MVrachev agree with this i noticed this too and commented just about the missing default breaking local runs. but the bigger thing i wasnt sure about was issue #258 explicitly says "use the dev version for all components" so i wasnt sure if pulling dev was intentional or not.  makes sense that build context should stay so ft actually tests the pr changes

yashasviyadav30: @MVrachev  @Sandipmandal25

To give some context on my intent—I actually looked at how `repository-service-tuf-api` handles its testing environments (where its compose files pull the pre-built `worker:dev`). Combining that pattern with the explicit requirement in Issue #258 (*"use the 'dev' version for all components"*), my goal was just to enforce strict cross-repo consistency by forcing the `dev` tag globally.

But you're completely right on the technical side. While pulling `dev` works great for external services, doing it *inside* the worker repo itself creates a blind spot and bypasses validating the actual PR changes during FTs. 

I'll revert the `build: context: .` removal across the compose files to ensure the FT pipeline isolates and tests the PR context correctly. Pushing the fix shortly!

Makefile
yashasviyadav30: Since I was forcing the `dev` image pull in the compose files, I added this to support local `make` runs without breaking. But since we are reverting back to the local build context for the compose files, I will revert this Makefile change as well!

yashasviyadav30:
@MVrachev Gentle Ping, all the review feedback has been incorporated. Would appreciate a re-review at your convenience. Thank you!

PR #892, opened 2026-04-30
fix: roll back partial metadata when bootstrap finalize fails

### LINKED ISSUE #248: Task:  bootstrap - persist metadata files in the end
### What is the task about?

While working on issue https://github.com/vmware/repository-service-tuf-api/issues/181 I noticed that if there is an exception in https://github.com/vmware/repository-service-tuf-worker/blob/5ac572a5d05558d915f3576e770490bc2e8bbd89/repository_service_tuf_worker/repository.py#L395 in the end of the function then the task is considered as failed, but we have already saved all metadata files.
I suggest we move the persisting of metadata files just before the end before creating the final response.

This will make sure that the next time we try to do bootstrap we can save the new files without any errors.

Also, it has the side benefit that it removes the confusion from the user because if a user sees that the bootstrap has failed and we successfully remove the lock, meaning the `api/v1/get/bootstrap` says the system is ready for bootstrap, but there are already existing metadata files it could be confusing for the user.

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

theycallmeaabie:
## What is the PR about?                                                                                                              

If `_bootstrap_finalize` throws an exception after already writing some files,                                                        
those files stay on disk. The next bootstrap attempt then fails on stale data,
and the system stays locked because `BOOTSTRAP` never got reset to `None`.                                                            

The fix adds a `delete()` method to `IStorage` (and both backends) so the                                                             
bootstrap paths can clean up after themselves. Any files written before the
failure get deleted on the way out. If a delete fails it's logged and we move                                                         
on, better to leave one file than to stop and leave everything. Either way,                                                          
`BOOTSTRAP` and `ROOT_SIGNING` are reset so a retry works without manual                                                              
intervention.                                                                                                                                                                                                                                                  

Closes #248                                                                                                                           
Relates to repository-service-tuf-api#181

Tests have been added for the bug fix.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -140,6 +140,7 @@ class MetadataRepository:
         self._expire_timedelta = timedelta(hours=self._hours_before_expire)
         self._timeout = int(app_settings.get("LOCK_TIMEOUT", 500.0))
         self._uses_succinct_roles: Optional[bool] = None
+        self._persist_tracking: Optional[List[str]] = None
 
     @property
     def _settings(self) -> Dynaconf:
@@ -329,6 +330,8 @@ class MetadataRepository:
 
         bytes_data = role.to_bytes(JSONSerializer())
         self._storage_backend.put(bytes_data, filename)
+        if self._persist_tracking is not None:
+            self._persist_tracking.append(filename)
         return filename
 
     def _is_expired(self, role: str) -> Optional[str]:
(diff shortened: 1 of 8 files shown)


repository_service_tuf_worker/repository.py
```diff
@@ -140,6 +140,7 @@ def __init__(self):
         self._expire_timedelta = timedelta(hours=self._hours_before_expire)
         self._timeout = int(app_settings.get("LOCK_TIMEOUT", 500.0))
         self._uses_succinct_roles: Optional[bool] = None
+        self._persist_tracking: Optional[List[str]] = None
```
MVrachev, line 143: The rollback list is stored as self._persist_tracking on the MetadataRepository instance, and _persist() checks self._persist_tracking on every call.

That means every _persist() in the process - not just the ones inside _bootstrap_finalize - will append to that list while a bootstrap is running. If anything else triggers a _persist() in parallel (a second Celery task, a retry, future code that calls _persist from a different flow), its filenames get tracked too, and on failure the rollback will delete() files it didn't write.

The tracking should be scoped to the bootstrap call itself - for example, by passing a tracker list down into _bootstrap_finalize / _bootstrap_online_roles / _persist, or using a context manager - instead of sharing it through self.

PR #894, opened 2026-04-30
Validate keys type and algorithm

YYuan02:
What is the PR about?
This PR implements Validate keys type and algorithm as described in issue https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/202.

Changes:

import the securesystemslib [KEY_FOR_TYPE_AND_SCHEME](https://github.com/secure-systems-lab/securesystemslib/blob/34e50cb059ea630e7a808dca7c1744b20f2d8ce4/securesystemslib/signer/__init__.py#L30) dict
Update the KEY_FOR_TYPE_AND_SCHEME by clearing its content with dict.clear() and instead adding our supported algorithms
During the bootstrap ceremony when loading the public keys stored in root.signed.keys we should do it with the [Key.from_dict](https://github.com/secure-systems-lab/securesystemslib/blob/34e50cb059ea630e7a808dca7c1744b20f2d8ce4/securesystemslib/signer/_key.py#L74) API call to use the validation inside securesystemslib.

### DIFF
--- a/repository_service_tuf_worker/repository.py
+++ b/repository_service_tuf_worker/repository.py
@@ -26,7 +26,9 @@ from securesystemslib.signer import (
     Signature,
     Signer,
     SigstoreKey,
+    SSlibKey
 )
+
 from tuf.api.exceptions import (
     BadVersionNumberError,
     RepositoryError,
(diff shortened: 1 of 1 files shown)


MVrachev:
@YYuan02 you have problems with failing.
Run `tox run -e lint` to see them and fix them.

PR #903, opened 2026-05-26
Eliminate S3 list_objects overhead via version cache and pre-warmup

ArkadiuszNitkaSWI:
# Performance Fix: S3 Version Cache for RSTUF Worker

---

## Problem

Each `add_artifacts` task processes a batch of files across multiple TUF metadata
roles — typically `root`, `targets`, `snapshot`, and one or more `bins-xx` delegation
shards (up to 9+ roles per task depending on the hash prefix distribution). For every
role it processes, the worker calls `AWSS3.get()` without a known version number in
order to load the current metadata.

The original implementation resolved the latest version with:

```python
filenames = awswrangler.s3.list_objects(
    path=f"s3://{bucket}/*.{role}.json",
    boto3_session=self._s3_session,
)
version = max(int(name.split(...)[-1].split(".")[0]) for name in filenames)
```

S3 does not support server-side suffix or glob filtering. The `*.{role}.json` pattern
causes `list_objects` to enumerate **every object in the bucket** and filter
client-side. Cost scales with total bucket size, not with the number of matching
objects.

| Bucket size | `list_objects` cost per call |
|---|---|
| ~250 objects | ~0.3 s |
| ~5 500 objects | 2.5–4 s |
| ~12 000 objects | 3–4 s |

With 9+ calls per task, at 12 000 objects this accounted for **~27 s of a ~33 s total
task latency** — roughly 82% of end-to-end time. KMS signing (the other obvious
candidate) contributed only ~1.4 s total (7 calls × ~0.2 s each).

The bucket grows monotonically: every task writes multiple versioned role files
(`N.rolename.json`) and they are never deleted. The problem worsens as the system is
used.

---

## Fix

Two targeted changes to `AWSS3` in `awss3.py` (108 lines added, 15 removed; no new
dependencies):

### 1. In-process version cache

A class-level dict `_version_cache: Dict[str, int]` maps `role_name → latest known
version`. The `get()` method reads from this cache instead of calling `list_objects`:

```python
cached_version = AWSS3._version_cache.get(role)
if cached_version is not None:
    version = cached_version
else:
    # Cache miss: scan once and store the result.
    ...
    AWSS3._version_cache[role] = version
```

The `put()` method updates the cache after every successful write so the next `get()`
for that role resolves the correct key without any S3 listing:

```python
parts = filename.split(".", 1)
if parts[0].isdigit():
    role_name = parts[1][:-5]   # strip ".json"
    AWSS3._version_cache[role_name] = int(parts[0])
```

### 2. Pre-warmup at worker startup

`_prewarm_version_cache()` is called once from `configure()` when the Celery worker
initialises. It scans the entire bucket a single time (`list_objects` with no glob,
prefix `s3://bucket/`), parses every `N.rolename.json` filename, and records the
highest version per role. All roles are in cache before the first task arrives, so
there is no cold-start penalty on the first job.

```
S3 version cache warmed: 35 roles, 12 241 objects scanned (3.74s)
```

If the prewarm fails (network error, permissions issue), a warning is logged and the
worker starts with an empty cache, falling back to the original per-role scan on first
access — identical to pre-fix behaviour.

---

## Benchmark Results

All runs: **10 concurrent jobs × 10 files each = 100 total artifacts**, real AWS S3
(`us-east-1`) + AWS KMS RSA-4096 online signing, Minikube on EC2 `t3.medium`.
`task_wait` = time from API POST to Celery `SUCCESS` (excludes S3 upload and HTTP
overhead).

| Benchmark | Condition | p50 | p95 | Wall (10 jobs) |
|---|---|---|---|---|
| **Bench5** | **Pre-fix baseline, ~12 000 bucket objects** | **36.6 s** | **42.4 s** | **6.3 min** |
| Bench6 | Version cache only, 1st run after restart | 11.8 s | 42.2 s | — |
| Bench7 | Version cache only, warm | 11.8 s | 22.1 s | — |
| Bench8 | Version cache only, fully warm | 9.9 s | 19.1 s | — |
| Bench9 | Cache + pre-warmup, instrumented build | 6.6 s | 8.7 s | 1.2 min |
| **Bench10** | **Cache + pre-warmup, clean build** | **5.6 s** | **8.7 s** | **1.1 min** |

**Speedup vs baseline: 6.5× p50, 4.9× p95, 5.7× wall time.**

The p95 spike in Bench6 (42.2 s) is a single cold-shard DB cache miss immediately
after restart with version-cache only and no prewarm; it disappears entirely once
pre-warmup is added (Bench9+).

### Remaining latency floor (~5–6 s)

| Component | Estimated cost |
|---|---|
| Celery task overhead + DB reads | ~2–3 s |
| KMS sign (7 calls × ~0.2 s) | ~1.4 s |
| S3 get/put (9+ calls × ~0.05 s) | ~0.5 s |
| **Total** | **~4–5 s** |

Consistent with the observed minimum of 5.6 s. No further reduction is possible
without addressing KMS latency or Celery/DB overhead.

---

### No new failure modes

- No external state (no Redis, no additional DB table, no shared memory).
- No new dependencies.
- A worker restart triggers a fresh prewarm (~4–8 s at 12 000 objects). The prewarm
  runs before the worker accepts tasks, so no task sees an empty cache unless the
  prewarm itself fails (handled gracefully).

### DIFF
--- a/repository_service_tuf_worker/services/storage/awss3.py
+++ b/repository_service_tuf_worker/services/storage/awss3.py
@@ -3,8 +3,6 @@
 #
 # SPDX-License-Identifier: MIT
 
-import logging
-import time
 from io import BytesIO
 from typing import Any, Dict, List, Optional
 
@@ -24,33 +22,11 @@ from repository_service_tuf_worker.interfaces import (
 
 
 class AWSS3(IStorage):
-    # In-process version cache: role_name -> latest known version number.
-    #
-    # AWSS3.get() needs to resolve the latest version of a role before it can
-    # fetch it from S3.  The original implementation called
-    # awswrangler.s3.list_objects(path="s3://bucket/*.{role}.json") on every
-    # call, which cannot use an S3 key-prefix scan and instead lists all
-    # objects in the bucket client-side.  At ~5 000+ objects that took 2.5–4 s
-    # per call; with 9+ calls per task it dominated task latency (~27 s of
-    # ~33 s total).
-    #
-    # The fix has two parts:
-    #
-    # 1. Pre-warmup (_prewarm_version_cache): called once per worker process
-    #    from configure() at startup.  Lists the entire bucket a single time,
-    #    parses every "N.rolename.json" filename, and stores the max version
-    #    seen for each role.  This runs before the first task arrives so there
-    #    is no cold-start penalty.
-    #
-    # 2. Cache maintenance: get() reads from the cache (skipping list_objects
-    #    entirely after warmup); put() updates the cache after every write so
-    #    the next get() fetches the correct version key directly.
-    #
-    # The cache is per-worker-process.  Celery ForkPoolWorkers each get their
-    # own copy after fork, independently pre-warmed.  A worker restart triggers
-    # a fresh prewarm (~4 s) before accepting tasks.
+    # Per-process version cache. Keys are role names, values are the
+    # highest version this process has seen. Acts as a lower bound: the
+    # real latest is always >= the cached value because a process only
+    # learns versions it wrote or explicitly probed.
     _version_cache: Dict[str, int] = {}
-    _cache_warmed: bool = False
 
     def __init__(
         self,
@@ -103,7 +79,7 @@ class AWSS3(IStorage):
             endpoint_url=endpoint,
         )
 
-        instance = cls(
+        return cls(
             bucket_name,
             s3_session,
             s3_client,
@@ -112,8 +88,6 @@ class AWSS3(IStorage):
             region,
             endpoint,
         )
-        cls._prewarm_version_cache(bucket_name, s3_session)
-        return instance
 
     @classmethod
     def settings(cls) -> List[ServiceSettings]:
@@ -144,50 +118,59 @@ class AWSS3(IStorage):
             ),
         ]
 
-    @classmethod
-    def _prewarm_version_cache(
-        cls, bucket: str, s3_session: boto3.Session
-    ) -> None:
-        """
-        Populate _version_cache with a single list_objects call at worker
-        startup.  Scans all objects in the bucket once, parses every
-        "N.rolename.json" filename, and records the highest version seen for
-        each role.  Subsequent get() calls skip list_objects entirely.
+    def _scan_latest_version(self, role: str) -> int:
+        """Full bucket scan to find the highest version for a role.
 
-        Guarded by _cache_warmed so it runs at most once per worker process
-        even if configure() is called multiple times.
+        This is the original list_objects path and is expensive at scale
+        (scans every object in the bucket client-side). It is used only
+        when the version cache has no entry for the role.
         """
-        if cls._cache_warmed:
-            return
-        t0 = time.perf_counter()
-        try:
-            all_keys = awswrangler.s3.list_objects(
-                path=f"s3://{bucket}/",
-                boto3_session=s3_session,
-            )
-        except Exception as e:
-            logging.warning("S3 version cache prewarm failed, skipping: %s", e)
-            return
-        prefix = f"s3://{bucket}/"
-        for s3_uri in all_keys:
-            filename = s3_uri[len(prefix):]
-            parts = filename.split(".", 1)
-            if (
-                len(parts) == 2
-                and parts[0].isdigit()
-                and filename.endswith(".json")
-            ):
-                role_name = parts[1][:-5]  # strip ".json"
-                version = int(parts[0])
-                if version > cls._version_cache.get(role_name, 0):
-                    cls._version_cache[role_name] = version
-        cls._cache_warmed = True
-        logging.info(
-            "S3 version cache warmed: %d roles, %d objects scanned (%.2fs)",
-            len(cls._version_cache),
-            len(all_keys),
-            time.perf_counter() - t0,
+        s3_path = f"s3://{self._bucket}/"
+        filenames = awswrangler.s3.list_objects(
+            path=f"{s3_path}*.{role}.json",
+            boto3_session=self._s3_session,
         )
+        versions = [
+            int(name.split(s3_path)[-1].split(".", 1)[0]) for name in filenames
+        ]
+        try:
+            return max(versions)
+        except ValueError:
+            return 1  # no existing file -> start at 1
+
+    def _latest_version(self, role: str) -> int:
+        """Return the latest version number for a role, self-healing across
+        processes.
+
+        The cached value is a lower bound: a process only ever learns versions
+        it wrote or listed, so its cache is always <= the real latest.
+
+        After obtaining the hint (from cache or a full scan), we probe upward
+        one step at a time using cheap head_object calls. TUF versions are
+        strictly contiguous (n -> n+1), so the loop terminates after at most
+        one probe in the steady state (when this process is the sole writer)
+        or a few probes when another process has advanced the version.
+        """
+        version = AWSS3._version_cache.get(role)
+        if version is None:
+            # Cache miss: fall back to the full bucket scan once.
+            version = self._scan_latest_version(role)
+
+        # Self-heal across processes: probe upward until the next key is
+        # absent. Because TUF versions are contiguous this converges quickly
+        # even when another Celery worker has already written ahead.
+        client = self._s3_client
+        while True:
+            try:
+                client.head_object(
+                    Bucket=self._bucket, Key=f"{version + 1}.{role}.json"
+                )
+                version += 1
+            except ClientError:
+                break
+
+        AWSS3._version_cache[role] = version
+        return version
 
     def get(self, role: str, version: Optional[int] = None) -> Metadata[T]:
         """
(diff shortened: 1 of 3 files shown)


repository_service_tuf_worker/services/storage/awss3.py
```diff
@@ -22,6 +24,34 @@
...
+    #    is no cold-start penalty.
+    #
+    # 2. Cache maintenance: get() reads from the cache (skipping list_objects
+    #    entirely after warmup); put() updates the cache after every write so
+    #    the next get() fetches the correct version key directly.
+    #
+    # The cache is per-worker-process.  Celery ForkPoolWorkers each get their
+    # own copy after fork, independently pre-warmed.  A worker restart triggers
+    # a fresh prewarm (~4 s) before accepting tasks.
+    _version_cache: Dict[str, int] = {}
```
srinjoydutta03: This comment describes ForkPoolWorker copy-on-write behaviour, but the cache is scoped per process. Any deployment with multiple worker processes  will hold independent `_version_cache` instances, each populated by its own `configure()` / prewarm. It introduces a risk for stale read scenario leading to inconsistencies.

Please refer to this file [supervisor.conf](https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/main/supervisor.conf) for such a scenario. The layout here is that
two Celery consumers are on separate queues, is one example of this topology.

Was multi-process behaviour tested under this layout?

srinjoydutta03 (changes requested):

srinjoydutta03:
Hi @ArkadiuszNitkaSWI it's very good finding and performance analysis, please look at the comment above and let me know your thoughts on it. If you can please a raise an issue and link it to the PR

repository_service_tuf_worker/services/storage/awss3.py
ArkadiuszNitkaSWI: No. The benchmarks ran with this exact supervisor.conf topology (single pod, two Celery processes via supervisord), but all benchmark jobs were submitted sequentially, one at a time, waiting for completion before the next. Concurrent multi-process cache interaction could be never exercised.

What would be the suggested approach to resolve this, should the database be treated as the single source of truth for role versions, with all `get()` call sites passing the version explicitly from rstuf_target_roles database?

ArkadiuszNitkaSWI:
Hello! Created the issue #908

repository_service_tuf_worker/services/storage/awss3.py
MVrachev: This implementation has a problem as rightly pointed by @srinjoydutta03.
Here is an example:

Imagine 2 child processes, both pre-warmed at startup when snapshot was v9, so both caches say {"snapshot": 9}.

1. Task N is dispatched to process A. A acquires `LOCK_TARGETS`. It reads get("snapshot") → cache says `9` → reads 9.snapshot.json, bumps to 10, put("10.snapshot.json"). A's cache is now `{"snapshot": 10}`. A releases the lock.
2. Task N+1 is dispatched to process B (B was idle, so the parent picks it). B acquires LOCK_TARGETS - the lock is free, A is done, so this is perfectly legal serialization. B calls get("snapshot") → **B's cache still says 9** 
A's write never touched B's memory).
3. B reads 9.snapshot.json, bumps to 10, and writes 10.snapshot.json again - overwriting A's snapshot, off a stale base.

It's a problem that is per process cache.

One solution that you can use is to add 3 methods with:

```python
def _latest_version(self, role: str) -> int:
    # Cached value is a lower bound: a process only ever learns versions it
    # wrote or listed, so its cache is always <= the real latest.
    version = AWSS3._version_cache.get(role)
    if version is None:
        version = self._scan_latest_version(role)  # old list_objects path
    # Self-heal across processes: TUF versions are contiguous (n -> n+1),
    # so probe upward until the next version is absent.
    while self._object_exists(f"{version + 1}.{role}.json"):
        version += 1
    AWSS3._version_cache[role] = version
    return version

def _object_exists(self, key: str) -> bool:
    try:
        self._s3_client.head_object(Bucket=self._bucket, Key=key)
        return True
    except ClientError:
        return False

def _scan_latest_version(self, role: str) -> int:
    s3_path = f"s3://{self._bucket}/"
    filenames = awswrangler.s3.list_objects(
        path=f"{s3_path}*.{role}.json",
        boto3_session=self._s3_session,
    )
    versions = [
        int(name.split(s3_path)[-1].split(".", 1)[0]) for name in filenames
    ]
    try:
        return max(versions)
    except ValueError:
        return 1  # no existing file -> start at 1
```


and the flow should be in the same case:


Starting state: both processes pre-warmed at v9, so A's cache and B's cache each say {"snapshot": 9}.

- Task N → process A

1. A acquires LOCK_TARGETS, calls get("snapshot").
2. _latest_version("snapshot"): hint = 9. Probe S3: does 10.snapshot.json exist? No. → latest = 9.
3. A reads 9.snapshot.json, bumps to 10, put("10.snapshot.json"). A's cache → {"snapshot": 10}. Releases lock.
So far identical to before - no difference yet, because A was the up-to-date process.

Task N+1 → process B (the case that broke before)

1. B acquires LOCK_TARGETS (free, legal).
2. B calls get("snapshot") → _latest_version("snapshot"): B's hint is still 9 (A's write never touched B's memory - that fact hasn't changed).
3. Here is the fix. Instead of trusting the hint, B probes S3: does 10.snapshot.json exist? Yes - A just wrote it. So B advances: hint → 10. Does 11.snapshot.json exist? No. → latest = 10. B updates its cache to {"snapshot": 10}.
4. B reads 10.snapshot.json (A's fresh write), bumps to 11, writes 11.snapshot.json.

The version chain is now ...9 → 10 (by A) → 11 (by B). No overwrite, no rollback. 
B built on top of A's work instead of clobbering it.


How does that sound?

@kairoaraujo give your opinion if you want as well.

ArkadiuszNitkaSWI: Hi @MVrachev, @srinjoydutta03 
 I implemented and tested your suggested solution. Performance is comparable to my original approach, but your design works in a multi-worker scenario,  One optional micro-optimisation worth considering, particularly for single-worker deployments: updating the cache inside put() after each write. This avoids the single head_object probe on the next _latest_version() call for that role, since the writer already knows the version it just wrote:
```

  parts = filename.split(".", 1)
  if len(parts) == 2 and parts[0].isdigit():
      role = parts[1].removesuffix(".json")
      written_version = int(parts[0])
      current = AWSS3._version_cache.get(role, 0)
      if written_version > current:
          AWSS3._version_cache[role] = written_version

```

kairoaraujo:
Hi @ArkadiuszNitkaSWI it looks very good improvement.
I'm reviewing it, as I'm planning to cut a new release it looks good to have.

repository_service_tuf_worker/services/storage/awss3.py
kairoaraujo: > Hi @MVrachev, @srinjoydutta03 I implemented and tested your suggested solution. Performance is comparable to my original approach, but your design works in a multi-worker scenario, One optional micro-optimisation worth considering, particularly for single-worker deployments: updating the cache inside put() after each write. This avoids the single head_object probe on the next _latest_version() call for that role, since the writer already knows the version it just wrote:
> 
> ```
> 
>   parts = filename.split(".", 1)
>   if len(parts) == 2 and parts[0].isdigit():
>       role = parts[1].removesuffix(".json")
>       written_version = int(parts[0])
>       current = AWSS3._version_cache.get(role, 0)
>       if written_version > current:
>           AWSS3._version_cache[role] = written_version
> ```

Maybe it could be a sub-issue for the opened issue, as a followup improvement.

tests/unit/tuf_repository_service_worker/services/storage/test_awss3.py
```diff
@@ -82,9 +82,7 @@ def test_configure(self, mocked_boto3):
                 endpoint_url=service._endpoint_url,
             )
         ]
-        assert awss3.boto3.Session().resource().buckets.all.calls == [
-            pretend.call()
-        ]
+        assert awss3.boto3.Session().resource().buckets.all.calls == [pretend.call()]  # noqa: E501
```
kairoaraujo:
```suggestion
        assert awss3.boto3.Session().resource().buckets.all.calls == [pretend.call()]  # noqa: E501
        assert awss3.boto3.Session().resource().buckets.all.calls == [
            pretend.call()
        ]        
```

tests/unit/tuf_repository_service_worker/services/storage/test_awss3.py
```diff
@@ -324,12 +329,12 @@ def test_get_max_version_ValueError(self, mocked_boto3, monkeypatch):
...
+        get_obj = pretend.call_recorder(lambda *a, **kw: fake_aws3_object)
         service._s3_client = pretend.stub(
-            get_object=pretend.call_recorder(lambda *a, **kw: fake_aws3_object)
-        )
-        monkeypatch.setitem(
-            awss3.__builtins__, "max", pretend.raiser(ValueError)
+            get_object=get_obj,
+            head_object=pretend.raiser(awss3.ClientError({}, "head_object")),
         )
+        monkeypatch.setitem(awss3.__builtins__, "max", pretend.raiser(ValueError))  # noqa: E501
```
kairoaraujo:
```suggestion
        monkeypatch.setitem(
            awss3.__builtins__, "max", pretend.raiser(ValueError)
        )        
```

repository_service_tuf_worker/services/storage/awss3.py
```diff
@@ -112,6 +118,60 @@ def settings(cls) -> List[ServiceSettings]:
...
+        # even when another Celery worker has already written ahead.
+        client = self._s3_client
+        while True:
+            try:
+                client.head_object(
+                    Bucket=self._bucket, Key=f"{version + 1}.{role}.json"
+                )
+                version += 1
+            except ClientError:
+                break
```
srinjoydutta03: @ArkadiuszNitkaSWI Thanks for the changes! It's great and It looks good to me. 

Just a nit: we should explicitly catch the not found error and raise other type of errors I think. Probably like this,

```
except ClientError as error:
    status_code = error.response.get("ResponseMetadata", {}).get("HTTPStatusCode")
    if status_code == 404:
        break
    raise
```

kairoaraujo (approved):

srinjoydutta03 (approved):
LGTM!

kairoaraujo:
Thank you @ArkadiuszNitkaSWI for this outstanding contribution! 🎉

ArkadiuszNitkaSWI:
@kairoaraujo @srinjoydutta03 @MVrachev thank you for the kind words and review! I'm glad I could help and contribute ;D

PR #921, opened 2026-07-03
Revert "build: bump python to 3.14 and use hash in Dockerfile to be updated by dependabot"

kairoaraujo:
Reverts repository-service-tuf/repository-service-tuf-worker#919
It was merged with FT errors.
I will re-open with fixes the affects the FT tests, it is caused by 3.14


kairoaraujo:
I will propose a fix straightway

Issue #927, opened 2026-07-08
Bug: Worker stops processing after PostgreSQL connection loss - PendingRollbackError blocks all subsequent tasks


rseleven:
### What happened?

**Environment:**
- Worker v1.0.0

**Description:**

When a Celery worker temporarily loses its connection to PostgreSQL (`server closed the connection unexpectedly`), the SQLAlchemy session enters an invalid state (`PendingRollbackError`). All subsequent tasks delivered to the same worker process keep failing with:

sqlalchemy.exc.PendingRollbackError: Can't reconnect until invalid transaction is rolled back.  Please rollback() fully before proceeding

The worker does not recover until it is restarted manually.

**Steps to reproduce:**
1. Start the worker with a PostgreSQL backend.
2. Trigger any DB-bound task (e.g. `remove_artifacts` or `bump_online_roles`).
3. While the worker is running, restart or temporarily block the PostgreSQL server.
4. Re-enable PostgreSQL and send a new task to the same worker process.

**Expected behaviour:**
The worker should recover from a transient DB connection failure - rollback the broken transaction and reconnect automatically - so that subsequent tasks succeed.

**Actual behaviour:**
Every new task on the affected `ForkPoolWorker` raises `PendingRollbackError` indefinitely until the worker is restarted.

log:
```
[2026-07-08 09:25:53,197: INFO/MainProcess] Task app.repository_service_tuf_worker[bd5ec6006f1844d4a33f65effa12a055] received
[2026-07-08 09:25:53,318: ERROR/ForkPoolWorker-15] Task app.repository_service_tuf_worker[bd5ec6006f1844d4a33f65effa12a055] raised unexpected: OperationalError('(psycopg2.OperationalError) server closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n')
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        cursor, str_statement, effective_parameters, context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/sqlalchemy/engine/default.py", line 951, in do_execute
    cursor.execute(statement, parameters)
[...]

PR #928, opened 2026-07-08
fix: add stale connection detection and recovery for DB pool connections

rseleven:
When PostgreSQL closes an idle connection (timeout, restart, network issue), the worker's long-lived SQLAlchemy session enters a pending rollback state and all subsequent tasks fail with PendingRollbackError.

- Add pool_pre_ping=True to validate connections on checkout from pool
- Add pool_recycle=1800 to auto-recycle idle connections after 30 min
- Enable TCP keepalives (idle=30s, interval=5s, count=3) so the OS detects dead sockets and closes them proactively
- Rollback session on OperationalError to restore clean state
- Add autoretry_for to automatically retry failed tasks up to 3 times with exponential backoff

Closes issue #927

### DIFF
--- a/repository_service_tuf_worker/models/__init__.py
+++ b/repository_service_tuf_worker/models/__init__.py
@@ -2,8 +2,11 @@
 # SPDX-FileCopyrightText: 2022-2023 VMware Inc
 #
 # SPDX-License-Identifier: MIT
+import os
+import sys
+
 from sqlalchemy import create_engine
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session, sessionmaker
 
 from repository_service_tuf_worker.models.targets import (  # noqa
     crud as targets_crud,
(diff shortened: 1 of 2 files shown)


repository_service_tuf_worker/models/__init__.py
```diff
@@ -17,7 +17,17 @@
...
+        db_server,
+        pool_pre_ping=True,
+        pool_recycle=1800,
+        connect_args={
+            "keepalives": 1,
+            "keepalives_idle": 30,
+            "keepalives_interval": 5,
+            "keepalives_count": 3,
+        },
+    )
```
kairoaraujo, line 44: Maybe we should make it also configured via env | helm charts values?

kairoaraujo (commented):
Hi @rseleven,

Thank you for the PR, I have got this issue before but I didn't went down to it. It is really great.

I made a single comment that is more a doubt.

repository_service_tuf_worker/models/__init__.py
rseleven: added:
RSTUF_DB_KEEPALIVE_IDLE
RSTUF_DB_KEEPALIVE_INTERVAL
RSTUF_DB_KEEPALIVE_COUNT

kairoaraujo: maybe after the release, also in the helm charts :D

kairoaraujo (approved):

repository_service_tuf_worker/models/__init__.py
rseleven: https://github.com/repository-service-tuf/helm-charts/pull/62

NikitaStasevich:
Hi!
We’ve run into the exact same issue with our infrastructure. I think we’re really lucky that someone is already working on it. Could you let me know when the fix is ​​expected to be merged? We’re really looking forward to it.

kairoaraujo:
> Hi! We’ve run into the exact same issue with our infrastructure. I think we’re really lucky that someone is already working on it. Could you let me know when the fix is ​​expected to be merged? We’re really looking forward to it.

I'm planning to merge it in the following days and cut a new patch release for worker

kairoaraujo (approved):

kairoaraujo:
Thanks @rseleven for this important contribution.

PR #952, opened 2026-08-06
fix: replace poisoned session recovery with scoped_session lifecycle

rseleven:
When PostgreSQL drops a connection, the worker's long-lived SQLAlchemy Session enters a PendingRollbackError state that never recovers. PR #928 addressed this at the connection pool level (pool_pre_ping, TCP keepalives) but only added autoretry_for to the main task repository_service_tuf_worker, not to _update_online_role or _update_snapshot_timestamp which are the tasks that actually fail in the bump_online_roles chain.

Changes:
- Replace static self._db Session with scoped_session + sessionmaker bound to the same engine, so each task gets a fresh session
- Add _reset_db_session() method that clears the scoped session registry via Session.remove(), disposing the connection entirely
- Add @signals.task_postrun.connect handler that calls _reset_db_session() after every Celery task, ensuring the next task always starts with a clean session regardless of prior failures
- Add autoretry_for=(OperationalError,) with max_retries=3 and exponential backoff to _update_online_role and _update_snapshot_timestamp
- Make _db a property with getter/setter to preserve backward compatibility with existing test mocks (test_repo._db = pretend.stub)

### DIFF
--- a/app.py
+++ b/app.py
@@ -136,7 +136,13 @@ def _end_bor_chain_callback(result, start_time: float):
     return {"result": result, "execution_time_seconds": total_time}
 
 
-@app.task(serializer="json", queue="rstuf_internals")
+@app.task(
+    serializer="json",
+    queue="rstuf_internals",
+    autoretry_for=(OperationalError,),
+    max_retries=3,
+    retry_backoff=True,
+)
 def _update_online_role(role: str) -> Optional[str]:
     """
     Update online role (DB and JSON)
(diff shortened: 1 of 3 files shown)


kairoaraujo (approved):
Great to see this fix! 🎉

rseleven:
fix: correct import sorting in repository.py

rseleven:
style: fix ruff-format violation in repository.py sessionmaker call

rseleven:
fix: mock rstuf_db in test_repo fixture to support get_bind

kairoaraujo (approved):

PR #961, opened 2026-08-23
fix: delegated role key removal never removed anything

amaydixit11:
## The bug

`_remove_delegated_role_keys` collects the keyids still in use by *other*
roles, then deletes the target role's keys that are not among them. The loop
variable is a role **name**, but it is compared against the `DelegatedRole`
object:

```python
for not_delete in targets.signed.delegations.roles:
    if not_delete == delegated:      # str == DelegatedRole, never true
        continue
    other_keys += targets.signed.delegations.roles[not_delete].keyids
```

So the `continue` never runs, the role being removed contributes its own
keyids to `other_keys`, every key looks like it is still used by another
role, and **nothing is ever deleted**.

## Why it matters

Key objects outlive the role that referenced them. Delete a delegation, or
update one so it no longer trusts a key, and that key stays in
`targets.signed.delegations.keys` with nothing pointing at it.

Clients ignore unreferenced keys, so this is untidy rather than unsafe. But it
grows without bound, and it makes the metadata harder to audit: a reader
cannot tell a key that is still trusted from one that was rotated out. That is
the case I hit — after replacing a delegated role's signing key, the old key
was still sitting in the metadata.

## The second bug it was hiding

The unreachable line called `.delete(key)`:

```python
targets.signed.delegations.keys.delete(key)
```

`delegations.keys` is a plain `dict` and has no `.delete()`. Fixing only the
comparison would have turned a silent no-op into an `AttributeError` on the
first real removal. This uses `.pop(key, None)` instead, which also means a
role referencing a keyid with no corresponding key object cannot raise here.

## Tests

Three new tests, all of which fail before this change:

| Test | Asserts |
|---|---|
| `test__remove_delegated_role_keys` | a key shared with another role is kept, an exclusive one is dropped |
| `test__remove_delegated_role_keys_sole_role` | with no other role holding them, all the role's keys go |
| `test__remove_delegated_role_keys_dangling_keyid` | a keyid with no key object is skipped rather than raising |

Failure output before the fix:

```
AssertionError: assert ['shared_key', 'only_role_1'] == ['shared_key']
  Left contains one more item: 'only_role_1'
AssertionError: assert {'key_1': 'ke...: 'key_2_obj'} == {}
  Left contains 2 more items
```

`238 passed, 1 skipped`. `flake8`, `isort` and `bandit` clean. `ruff-format`
reports the same diff count against this branch as against `main` (1369 lines
either way), so no new formatting drift is introduced.

## Note on scope

This is independent of the delegation work I have open elsewhere; it reproduces
on `main` unchanged, which is why it is a standalone PR.


amaydixit11:
Closing to re-open this stacked on top of #950 in my fork instead, so it reviews as part of that series rather than as a separate change against main. Same commit, rebased onto the feature branch.

